Issue #96, opened 2016-06-20
Instead of “changed X” say “added X” if X was added


neongreen:
Currently we have this, it's silly:

![image](https://cloud.githubusercontent.com/assets/1523306/16192534/8c82cf06-36f2-11e6-80b4-b4fb23179a48.png)

How to fix it:

* in src/View.hs, in `renderEdit` check for emptiness in branches like `Edit'SetCategoryNotes`, etc. and render them differently if they are empty

babhishek21:
I believe the file `View.hs` resides in `/lib`and not `/src`.

neongreen:
Yeah, it was moved together with the rest of `/src`.

babhishek21:
So are we supposed to check for all the `Edit'Set*` cases to see if the old value is empty? What about when new values are empty?

neongreen:
> So are we supposed to check for all the Edit'Set* cases to see if the old value is empty?

Yeah. I wish I had a less boilerplatey definition of `Edit` (with some kind of first-class record fields), but currently it is what it is, and I don't know how to make it better without losing type safety. If you have any ideas, please tell!

> What about when new values are empty?

They can't be, because code that generates `Edit`s won't create an edit when the old value is equal to the new value. Furthermore, even if an empty value slips through, nothing bad will happen (the admin interface isn't even visible for ordinary users), so you can ignore this possibility.

Here's the function which adds an edit: https://github.com/aelve/guide/blob/master/lib/Guide.hs#L180-L190. Here's the function which checks that an edit actually changes something: https://github.com/aelve/guide/blob/master/lib/Types.hs#L672-L711.

babhishek21:
Great! I think I'll take a look today evening. Does anyone mind?

neongreen:
I definitely don't!

neongreen:
Fixed by @babhishek21 in #131 and deployed.

PR #131, opened 2016-10-16
Quick-fix for Edit'Set* cases when old Text value is empty

babhishek21:
Quick Fix to issue #96.

+ Currently, when old `Text` type value is empty, the descriptive text uses the word _"added"_ instead of _"changed"_. Other than that, the table column for the empty old value still exists.
(See image below for example)
image
Let me know if empty value columns require removal. That will require some more ninja chops. :hocho: 

+ Only some of the cases matching `Edit'Set*` have been modified. The ones which have not been touched are those which either have default old values which are not empty or `null`; _or_ have a default enumerated value assigned to them (system-defined or `Nothing`), at the time of creation.

==

I am an absolute _noob_ when it comes to Haskell. 
+ I hope my use of the `if-then-else` construct was justified (it seems to get the desired effect).
+ When I needed to check if a `Data.Text` type object was empty, I referred to https://github.com/aelve/guide/blob/master/lib/View.hs#L62-L64 and only `oldValue == T.empty` or 
`(T.null oldValue)` seemed to work. I used the latter. (I don't fully understand why `oldValue == Text.empty` or `(Text.null oldValue)` don't work. I'll leave that as is.)

### DIFF
--- a/lib/View.hs
+++ b/lib/View.hs
@@ -413,7 +413,8 @@ renderEdit globalState edit = do
     Edit'SetCategoryNotes catId oldNotes newNotes -> do
       p_ $ (if (T.null oldNotes) then "added" else "changed") >> " notes of category " >> printCategory catId
       table_ $ tr_ $ do
-        td_ $ blockquote_ $ toHtml (toMarkdownBlock oldNotes)
+        unless (T.null oldNotes) $
+          td_ $ blockquote_ $ toHtml (toMarkdownBlock oldNotes)
         td_ $ blockquote_ $ toHtml (toMarkdownBlock newNotes)
     Edit'SetCategoryProsConsEnabled catId _oldVal newVal -> do
       if newVal == True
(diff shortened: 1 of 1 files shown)


neongreen:
> I hope my use of the if-then-else construct was justified

Yep.

> I don't fully understand why `oldValue == Text.empty` or `(Text.null oldValue)` don't work

`Text.null` doesn't work because of how Haskell import syntax works. In `T.null`, `T` is the module alias and `null` is the name of the function. Now:

* When `Data.Text` is imported like `import qualified Data.Text as T`, it makes `T` refer to `Data.Text`.
* If `Data.Text` was imported like `import Data.Text as T`, it'd make `null` available both as `null` and as `T.null`. People don't usually import `Data.Text` like that because lots of names there clash with ones in Prelude.
* If `Data.Text` was imported plainly like `import Data.Text`, it'd only make `Data.Text.null` available as `null` and it'd be impossible to call because it clashes with `Prelude.null`.

(By the way, you don't need parens around `(T.null oldNotes)` because `if` is a keyword, not a function, and so Haskell can figure out unambiguously that `T.null oldNotes` is the condition here.)

> oldValue == T.empty

`T.null oldValue` is the preferred version. Generally in Haskell you use pattern-matching, `case` or predicates like `null` and `isNothing` when possible, though `foo == "blah"` is still more acceptable than `foo == ""` – this has to do with the fact that when you use `foo == []` on lists in general, it works in less cases than `null foo` (because e.g. you can't compare an `[Int -> Int]` list with anything but you can still check whether it's empty or not). There are also other arguments here: http://stackoverflow.com/questions/3853922/null-instead-of.

> Other than that, the table column for the empty old value still exists.

Can you fix this as well? It's not hard, you only need to conditionally execute the code that adds the table cell – e.g. 

```haskell
unless (T.null oldEcosystem) $ 
  td_ $ blockquote_ $ toHtml (toMarkdownBlock oldEcosystem)
```

neongreen:
(Well, and check that CSS doesn't get broken when there's only one cell.)

babhishek21:
>> Other than that, the table column for the empty old value still exists.
>
> Can you fix this as well?

Done. :smile: 

> (Well, and check that CSS doesn't get broken when there's only one cell.)

The CSS did not break. In fact it looks the same as the case when you add a **pro** or a **con**. This is what it looks like:
image

PR #141, opened 2017-01-24
Fix some lints

AaronFriel:
These just make `hlint` and `ghc-mod` happier. There was a lint I did not change (redundant typeclass) which shows up on a `-Wall`.

### DIFF
--- a/lib/Cache.hs
+++ b/lib/Cache.hs
@@ -1,7 +1,4 @@
-{-# LANGUAGE
-NoImplicitPrelude
-  #-}
-
+{-# LANGUAGE NoImplicitPrelude #-}
 
 module Cache
 (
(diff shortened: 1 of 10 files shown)


AaronFriel:
It appears the language pragma issue is this: https://github.com/ndmitchell/hlint/issues/125 

```
{-# LANGUAGE
  OverloadedStrings,
  ScopedTypeVariables,
  TypeFamilies,
  DataKinds,
  FlexibleContexts
  #-}
```

Parsed fine.

neongreen:
* I'm fine with changing pragmas style, it makes things easier if you use e.g. Emacs's automatic pragma insertion. I wanted to do it myself.

* I think I had some subtle and vague reasons for why `when (not ...)` was preferred to `unless` in some places, but I guess it does more harm than good.

PR #155, opened 2017-04-23
[GD-9] Restyle item-info form

juanbono:


templates/item-info.widget
```diff
@@ -152,64 +152,84 @@ function editItemInfo(itemUid) {
...
+  margin-top: 15px;
+  margin-bottom: 5px;
+}
+
+.form-btn-group {
+  margin-top: 20px;
+  margin-bottom: 5px;
+}
+
+.save {
```
neongreen: I think you should rename this class because it will clash with classes elsewhere.

(All CSS from templates is concatenated into one file.)

templates/item-info.widget
```diff
@@ -152,64 +152,84 @@ function editItemInfo(itemUid) {
...
 HTML: item-info-edit-form
 ------------------------------------------------------------
 {{! "autocomplete=off" everywhere: http://stackoverflow.com/q/8311455 }}
-<form onsubmit="submitItemInfo('{{item.uid}}', this); return false;">
-  <label>
-    Name<br>
-    <input name="name" value="{{item.name}}"
-           type="text" autocomplete="off">
+<form class="form-group" onsubmit="submitItemInfo('{{item.uid}}', this); return false;">
+  <label class="formLabel" for="name">
```
neongreen: If all labels have class `formLabel`, isn't it easier to just add a style for e.g. `.item-info-edit-form > label`?

PR #156, opened 2017-04-23
[GD-10] Show category description

juanbono:


src/Guide/Views.hs
```diff
@@ -701,6 +700,8 @@ renderSearchResult r = do
       SRCategory cat -> do
         a_ [class_ "category-link", href_ (categoryLink cat)] $
           toHtml (cat^.title)
+        div_ [class_ "category-description notes-like"] $
+          toHtml (extractSummary $ cat^.notes^.mdMarkdown) -----
```
neongreen: Why `-----`?

templates/item-info.widget
```diff
@@ -152,64 +152,84 @@ function editItemInfo(itemUid) {
   switchSection("#item-" + itemUid + " .item-info", "editing");
 }
 
+CSS
```
neongreen: This shouldn't be in this PR. 

(You can remove these changes by amending the commit with `templates/item-info.widget` unstaged and force-pushing.)

src/Guide/Markdown.hs
```diff
@@ -139,6 +140,10 @@ stringify = T.concat . map go
       HTML_BLOCK _      -> ""
       HTML_INLINE _     -> ""
 
+-- | Extract the first paragraph.
+extractSummary :: [MD.Node] -> MarkdownBlock
```
neongreen: This will extract the first heading if there isn't a paragraph preceding it. Instead you should parse Markdown with `toMarkdownBlockWithTOC` and extract all content before the first section ([`preface`](http://hackage.haskell.org/package/cmark-sections-0.1.0.3/docs/CMark-Sections.html#v:preface) from `mdTree`).

PR #159, opened 2017-05-22
Add category link

v0d1ch:
I never seen Spock so I just got the whole thing to compile and I guess it should be fine. That is why I love Haskell :)

### DIFF
--- a/src/Guide/Handlers.hs
+++ b/src/Guide/Handlers.hs
@@ -243,14 +243,14 @@ setMethods = Spock.subcomponent "set" $ do
           ("modified" :: Text, modified),
           ("merged" :: Text, merge original content' modified)]
   -- Trait
-  Spock.post (itemVar <//> traitVar) $ \itemId traitId -> do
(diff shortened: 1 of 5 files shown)


v0d1ch:
I see this fails due to the saved checkpoint ? I remember clicking on the button and I did not import the remote database just fyi
```Could not parse saved checkpoint due to the following error: Failed reading: safecopy: <unkown type>: Cannot find getter associated with this version number: Version {unVersion = 0}

From:	Guide.State.GlobalState:

	Guide.Types.Action.Action:

	Guide.Types.Edit.Edit:

CallStack (from HasCallStack):

  error, called at src/Data/Acid/Local.hs:322:7 in acid-state-0.14.2-4srzFjtJIJJHz8b7WcPoy9:Data.Acid.Local```

neongreen:
> I see this fails due to the saved checkpoint?

This is because you haven't added migration instances

neongreen:
I'm going to merge this PR after I do some changes. Notably, `categoryVar` doesn't need to be passed in some places because item IDs are unique even between categories.

Issue #124, opened 2016-08-13
A way to find broken links 


neongreen:
Would be even better if all links were submitted to archive.org automatically.

neongreen:
https://www.notion.so/aelve/Broken-link-detection-and-archival-16e0ad91f8f446dcb5a9673d2b4ef0c4

PR #175, opened 2017-07-20
[WIP] [#124] Add admin/links with broken links

neongreen:
I'm going to create a PR out of this so that commenting would be easier.

### DIFF
--- a/src/Guide/Views.hs
+++ b/src/Guide/Views.hs
@@ -54,7 +54,8 @@ import Data.Time.Format.Human
 import qualified Data.Aeson as A
 -- CMark
 import qualified CMark as MD
-import CMark.Sections (annValue, flattenDocument)
+-- Generic traversal (for finding links in content)
+import Data.Generics.Uniplate.Data (universeBi)
 
 import Guide.Config
 import Guide.State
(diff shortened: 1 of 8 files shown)


src/Guide/Views.hs
```diff
@@ -42,11 +43,15 @@ import NeatInterpolation
 import Lucid hiding (for_)
 -- Network
 import Data.IP
+import Network.HTTP.Conduit
```
neongreen: Why conduit if you're only using things that `Network.HTTP.Conduit` reexports from http-client? Also, if you use `http-client-tls` instead you'll be able to query `https` links as well.

src/Guide/Views.hs
```diff
@@ -793,3 +798,81 @@ on those <div>s.
...
+    ++ concatMap findLinksTrait _itemCons
+    ++ findLinksMdBlock _itemEcosystem
+    ++ findLinksMdTree _itemNotes
+    ++ findLinksItemLink _itemLink
+
+  findLinksMdBlock :: MarkdownBlock -> [MD.NodeType]
+  findLinksMdBlock = findAllLinks . markdownBlockMdMarkdown
+
+  findLinksMdTree :: MarkdownTree -> [MD.NodeType]
+  findLinksMdTree = findAllLinks . parseMD . markdownTreeMdText
```
neongreen: `parseMD` will re-parse already parsed Markdown, it's better to use `annValue . flattenDocument . markdownTreeMdTree` (the former two functions come from `CMark.Sections`).

src/Guide/Views.hs
```diff
@@ -793,3 +798,81 @@ on those <div>s.
...
+    includeCSS "/markup.css"
+    includeCSS "/admin.css"
+    includeCSS "/loader.css"
+    title_ "Links – Aelve Guide"
+    meta_ [name_ "viewport",
+           content_ "width=device-width, initial-scale=1.0, user-scalable=yes"]
+
+  body_ $ do
+    script_ $ fromJS $ JS.createAjaxIndicator ()
+    h1_ "Links"
```
neongreen: Oops, I just noticed that non-broken links go into the table as well. Nevermind then :)

neongreen:
TODO:

* sort links into several tables (broken, not-broken, unknown)
* store link statuses in some kind of cache – e.g. a quick and dirty global IORef – and add a “re-check links” button to the page (because checking links isn't very cheap)
* say where each link comes from (at least which category/item)

src/Guide/Views.hs
```diff
@@ -793,3 +798,81 @@ on those <div>s.
...
+                  td_ "??"
+ where
+  handleHttpException :: HttpException -> String
+  handleHttpException (HttpExceptionRequest _ x) = show x
+  handleHttpException (InvalidUrlException _ x) = x
+
+  findLinksCategory :: Category -> [MD.NodeType]
+  findLinksCategory = filter isLink . findNodesCategory
+
+  isLink :: MD.NodeType -> Bool
```
neongreen: > so this simplification can be done later.

or not (because we want to track where links come from, and uniplate won't do that for us)

src/Guide/Views.hs
```diff
@@ -36,21 +36,25 @@ import Guide.Views.Category as X
 
 import Imports
 
+import Data.Monoid ((<>))
```
neongreen: Hmmm, pretty sure it should be available from `Imports`.

neongreen:
Looks good so far; haven't tried running it locally yet, but will do it tomorrow and probably approve

PR #176, opened 2017-07-25
Add support for multi-snippets with pretty renderer

vrom911:


neongreen:
I'm going to leave the branch around just in case, it does no harm.

PR #180, opened 2017-07-29
Support tables in markdown

vrom911:


src/Guide/Markdown.hs
```diff
@@ -366,3 +366,107 @@ instance SafeCopy MarkdownTree where
...
+   - smth 2
+   - smth 3
+@
+or
+@
+smth 1 | smth 2 | smth 3
+@
+These two examples are equal.
+-}
+getCellFromItem :: [MD.Node] -> Maybe [Text]
```
neongreen: * <code>getCell<b>s</b>FromItem</code>, since there are many returned cells. Maybe just `getCells`, actually.

* A cell should be `[MD.Node]` and not `Text`, since this way users will be able to use markup in table cells. (Breaking on `|` becomes a bit more effort this way, but still fairly easy to do. `*foo* | **bar** | baz` should parse as three cells – [*foo*, **bar**, baz].)

* Also you can't assume that there'll be a paragraph in the list item – if the user tries to add a list or a code block in a table cell, there won't be a paragraph. I recommend http://spec.commonmark.org/dingus/ to quickly check how a piece of Markdown is parsed. If you take this:

   ````
   - ```
     code example
     in a table
     ```
   - some text
   ````

  you will get this AST, without a paragraph in the first item:

  ```xml
  <list type="bullet" tight="true">
    <item>
      <code_block>code example
  in a table
  </code_block>
    </item>
    <item>
      <paragraph>
        <text>some text</text>
      </paragraph>
    </item>
  </list>
   ```

* Finally, this function doesn't even handle lists currently, only the `|` syntax (despite what its documentation suggests).

src/Guide/Markdown.hs
```diff
@@ -366,3 +366,107 @@ instance SafeCopy MarkdownTree where
...
+{-|
+Possible row syntax  is
+@
++  - smth 1
+   - smth 2
+   - smth 3
+@
+or
+@
+smth 1 | smth 2 | smth 3
```
neongreen: Typo: `+ smth 1 | smth 2 | smth 3`, not `smth 1 | smth 2 | smth 3`

tests/MdTablesSpec.hs
```diff
@@ -0,0 +1,238 @@
...
+                    ]
+
+rowWithSeparatorMD :: [[MD.Node]]
+rowWithSeparatorMD = [ [ MD.Node Nothing (MD.TEXT "Another foo ") [] ]
+                     , [ MD.Node Nothing (MD.TEXT " Another bar ") [] ]
+                     , [ MD.Node Nothing (MD.TEXT " Another baz") [] ]
+                     ]
+
+rowOfListSeveralNodesMD :: Int -> [[MD.Node]]
+rowOfListSeveralNodesMD n =
```
neongreen: Can you explain why posinfo is calculated here but not elsewhere?

src/Guide/Markdown.hs
```diff
@@ -366,3 +379,226 @@ instance SafeCopy MarkdownTree where
...
+   - smth 2
+   - smth 3
+@
+or
+@
++  smth 1 | smth 2 | smth 3
+@
+These two examples are equal.
+-}
+getCells :: [MD.Node] -> Maybe [[MD.Node]]
```
vrom911: I believe it works the same way.

src/Guide/Markdown.hs
```diff
@@ -366,3 +379,226 @@ instance SafeCopy MarkdownTree where
...
+isBreak :: [MD.Node] -> Bool
+isBreak [MD.ThematicBreak_] = True
+isBreak _                   = False
+
+-- | Generates 'HTML' table from 'Table' structure
+renderTable :: (Monad m) => MarkdownTable -> HtmlT m ()
+renderTable MarkdownTable{..} = do
+  h3_ $ toHtml markdownTableName
+  table_ [class_ "sortable"] $ do
+    whenJust markdownTableColumns $ \clmns ->
```
vrom911: Oops..

tests/MdTablesSpec.hs
vrom911: It's specialty of the test example. Manipulations with lines in Markdown makes positions changed, and `==` in `shoulBe` won't work without certain position specifying.

neongreen:
Left to do:

* [ ] Add table rendering to the main Markdown pipeline.
* [ ] Add tests checking that section-ing Markdown that contains tables works correctly.
* [ ] Actually check it by hand, too.

neongreen:
Closing so that it wouldn't clutter the list of PRs. The branch should still linger around.

Issue #181, opened 2017-07-30
Fix `ghc` warnings


sectore:
- [x] Spock warnings about `Spock.subcomponent`:  
```Deprecated: "Subcomponents will be removed in the next major release. They break route rendering and should not be used. Consider creating helper functions for reusable route components"```
- [x] Misc. warnings about shadowing variables:  For example
```This binding for `sess_id' shadows the existing binding
      defined at src/Guide/Types/Session.hs:63:1```

PR #182, opened 2017-07-30
Fix ghc warnings (#181)

sectore:
It fixes `ghc` warnings as described in #181 

In general: What about to have more restrict imports of modules to avoid shadowing variables? Another problem could be the [Imports](https://github.com/aelve/guide/blob/master/src/Imports.hs) module to use it everywhere. With that we might import content we don't need in other modules, but we have to care about shadowing variables. 

Anyway, if these warnings are not important we could change `ghc` warning options to prevent its output. But I would suggest that everyone focus more on all warnings with the goal to compile `Guide` w/o any warnings :)

### DIFF
--- a/guide.cabal
+++ b/guide.cabal
@@ -75,6 +75,7 @@ library
       Guide.Views.Utils
         Guide.Views.Utils.Input
     Guide.Cache
+    Guide.Routes
   other-modules:
     Imports
   build-depends:       Spock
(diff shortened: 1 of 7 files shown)


neongreen:
> In general: What about to have more restrict imports of modules to avoid shadowing variables?

Well, sure, we can do that occasionally via qualified imports. I would still prefer things like `view` to always mean the `view` from lenses, though.

PR #189, opened 2017-08-14
Add archive.org data to links page

vrom911:


src/Guide/Views.hs
```diff
@@ -913,3 +948,35 @@ findLinksItem item = findLinksMD item' ++ maybeToList (item^.link)
 -- | Find all Markdown links in /any/ structure, using generics.
 findLinksMD :: Data a => a -> [Url]
 findLinksMD a = [url | MD.LINK url _ <- universeBi a]
+
+data ArchiveOrgResponse =
+  ArchiveOrgResponse { archivedSnapshots :: ArchivedSnapshot
+                     } deriving (Show, Generic)
+
+data ArchivedSnapshot =
```
neongreen: I think that you don't really need all those datatypes here, because the archive.org API is pretty poor and we only need to get one thing out of it:

``` haskell
data ArchivedSnapshot {
  snapAvailable :: Bool,
  snapUrl       :: Url,
  snapTimestamp :: UTCTime,
  snapStatus    :: Text }

instance FromJSON ArchivedSnapshot where
  parseJSON = withObject "ArchivedSnapshot" $ \o -> do
    snapAvailable <- o .: "available"
    snapUrl       <- o .: "url"
    snapTimestamp <- o .: "timestamp" >>=
                     parseTimeM True defaultTimeLocale "%Y%m%d%H%M%S"
    snapStatus    <- o .: "status"
    pure ArchivedSnapshot{..}
```

Btw, with `parseTimeM` we won't `error` out if we can't parse the timestamp, which is nice.

Then you need to parse the JSON with a manual parser, which is quite easy once you have a utility function for it (which for whatever reason isn't included into `aeson` proper). I'll go and add one to `Guide.Utils` myself soon and then update this PR.

src/Guide/Views.hs
```diff
@@ -838,56 +839,90 @@ renderAdminLinks globalState = do
...
+                td_ dt
+                td_ $ button "To Archive" [] (JS.saveToArchiveOrg [JS.toJS lnk])
  where
   handleHttpException :: HttpException -> LinkStatus
   handleHttpException (HttpExceptionRequest _ x) = Broken $ show x
   handleHttpException (InvalidUrlException  _ x) = Broken x
 
-  sortLinks :: [(a, b, LinkStatus)] -> ([(a, b)], [(a, b)], [(a, b, String)])
+  handleHttpExceptionDecode :: HttpException -> IO (String, String)
+  handleHttpExceptionDecode _ = pure ("error", "error")
```
neongreen: Can you explain what it does and why it returns two `"error"`s?

src/Guide/Views.hs
```diff
@@ -838,56 +839,90 @@ renderAdminLinks globalState = do
...
+
+  sortLinks :: [(a, b, LinkStatus, c)] -> ([(a, b, c)], [(a, b)], [(a, b, String, c)])
   sortLinks = foldr sortLink ([], [], [])
 
-  sortLink (a, b, OK)          = (\(x, y, z) -> ((a, b):x, y, z))
-  sortLink (a, b, Unparseable) = (\(x, y, z) -> (x, (a, b):y, z))
-  sortLink (a, b, Broken text') = (\(x, y, z) -> (x, y, (a, b, text'):z))
+  sortLink (a, b, OK, c)           = \ (x, y, z) -> ((a, b, c) : x, y, z)
```
neongreen: I think that once we are doing four-tuples, it's better to create a proper data type in this module and use it instead.

src/Guide/Views.hs
```diff
@@ -838,56 +839,90 @@ renderAdminLinks globalState = do
...
+                td_ location
+                td_ $ a_ [href_ lnk] (toHtml lnk)
       h2_ "OK Links"
       table_ [class_ "sortable"] $ do
           thead_ $ tr_ $ do
             th_ [class_ "sorttable_nosort"] "Category"
             th_ [class_ "sorttable_nosort"] "Link"
-          tbody_ $ do
-            for_ ok $ \(cat, l) -> do
+            th_ [class_ "sorttable_nosort"] "Latest archieve date"
```
neongreen: archieve → archiving

src/Guide/Views.hs
vrom911: I didn't perform this transformation on data types. I will wait for the JSON parser appear. 
Though I replaced time parser by `parseTimeM` .

src/Guide/Views.hs
vrom911: Replaced `"error"` by more informative messages.

PR #192, opened 2017-08-24
Add PublicDB type

vrom911:


src/Guide/Main.hs
```diff
@@ -155,6 +155,14 @@ mainWith config = do
        putStrLn "loaded the database successfully"
        closeAcidState db
        exitSuccess
+     when (args == ["--from-publicDB"]) $ do
+       db :: DB <- openLocalStateFrom "state/" (error "couldn't load state")
+       gs <- query db GetGlobalState
```
neongreen: Well, this doesn't load from public DB, this creates a public DB :) The logic I had in mind was like this: if you provide `--load-public <filename>`, it will load PublicDB from that file, convert it to GlobalState, and use that state from now on (regardless of whether or not `state/` exists; if it exists it should be replaced).

Issue #193, opened 2017-08-25
Provide `Express` server


sectore:
As [discussed](https://aelve.slack.com/archives/C3UQX6F45/p1503561056000257) with @neongreen we are going to provide a simple [Express](http://expressjs.com/) server to serve each page of the front-end. Going this way we are able to develop page by page w/o the need to wait of developing all content of a SPA. In the future we will move all pages back into a SPA.

PR #194, opened 2017-08-25
`Express` server + multi-pages Pux apps (#193)

sectore:
This is the start of re-writing the front-end page by page. First step is to provide a `Express` server (see `front-ps/server/Main.purs`) to serve each page. Each page will have a single `JS` entry point (which is a single `JS` file compiled from `PS`). This `JS` will be included by `Express` in a page. ~so that we can render different pages with the same `EJS` template (see `front-ps/public/views/layout.ejs`)~

_UPDATE (26th August):_ 
* Extract modules into folders `common|client|server`
* Add `Pux` example app of `category-detail` page (still in progress)
* Use `webpack` (for bundling client-side only)

_UPDATE 27th August:_
* Add server-side rendering for category-detail page
* Add `Pux` example app for category-overview (incl. server-side rendering)
* Add routes to Pux app
* Use `webpack` for bundling server-side as well


front-ps/README.md
```diff
@@ -36,14 +36,9 @@ yarn
 yarn build:server
```
vrom911: Are `npm` commands similar to these ones with `yarn`?

sectore: @vrom911 With 5d5fd84 I've updated `README` for using `npm` commands.

PR #206, opened 2017-09-14
Add Github auth

neongreen:


src/Guide/Auth.hs
```diff
@@ -0,0 +1,45 @@
...
+import           Imports
+
+import           Guide.App
+import           Guide.Auth.OAuth2
+-- import           Guide.Config
+-- import           Guide.Config.OAuth2
+-- import           Guide.Routes
+-- import           Guide.ServerStuff
+
+mkAuthApi :: Path '[] 'Open -> GuideM ctx ()
```
neongreen, line 44: Can you add a haddock comment? E.g.

```
-- | This function adds handlers for all authentication routes that
-- we need to support. Assuming that the prefix is @auth@, 'mkAuthApi'
-- will add handlers for these routes:
--
-- @
-- /auth/oauth2/github/...
-- /auth/oauth2/google/...
-- /auth/oauth2/generic/...
-- @
--
-- For details on the created routes, see 'mkOAuth2Api' and [...].
```

src/Guide/Config/OAuth2/GitHub.hs
```diff
@@ -0,0 +1,73 @@
+{- |
+Module      :  Guide.Config.OAuth2.GitHub
+Description :  GitHub OAuth2 authentication parameters & workflow.
+Copyright   :  (c) Aaron Friel
+License     :  BSD-3
+
+Maintainer  :  Aaron Friel <mayreply@aaronfriel.com>
+Stability   :  unstable | experimental | provisional | stable | frozen
```
neongreen: Forgot to pick one

src/Guide/Config/OAuth2/GitHub.hs
```diff
@@ -0,0 +1,73 @@
...
+          _endpointAccessToken  <- o .:? "access-token-url" .!= (def' ^. endpointAccessToken)
+          _endpointCallback     <- o .:? "callback-url"     .!= (def' ^. endpointCallback)
+          return $ OAuth2Endpoint {..}
+    return $ GitHubEndpoint {..}
+
+instance ToJSON GitHubEndpoint where
+  toJSON GitHubEndpoint{..} =
+    Object base
+    where
+      (Object base) = toJSON _githubBaseConfig
```
neongreen: :thinking:

src/Guide/Config/OAuth2/Google.hs
```diff
@@ -0,0 +1,54 @@
...
+
+-- | Google authentication details
+data GoogleEndpoint = GoogleEndpoint {
+  _googleEndpointConfig :: OAuth2Endpoint
+  }
+  deriving (Eq, Show)
+
+makeClassy ''GoogleEndpoint
+
+-- instance OAuth2Default GoogleEndpoint where
```
neongreen: Can you explain why this (and a lot of other stuff) is commented out? Is it because it doesn't work yet, or for some other reason?

src/Guide/State.hs
```diff
@@ -860,6 +875,38 @@ logoutUserGlobally key = do
...
+addCreds user userCreds = do
+  allCreds <- concat . toList <$> use creds
+  if canAddCreds userCreds allCreds
+  then do
+    creds %= M.insertWith (++) (user ^. userID) [userCreds]
+    return True
+  else
+    return False
+
+-- | Remove external credentials to a user account.
```
neongreen: “from”

src/Guide/State.hs
```diff
@@ -860,6 +875,38 @@ logoutUserGlobally key = do
...
+
+-- | Get user from creds
+loginUserCreds :: Creds -> Acid.Query GlobalState (Maybe User)
+loginUserCreds userCreds = do
+  matches <- filter (\(_, cs) -> any matchCred cs) . M.toList <$> view creds
+  case matches of
+    [(key, _)] -> view (users . at key)
+    _ -> return Nothing
+  where
+    matchCred cred =    cred ^. credsProvider == userCreds ^. credsProvider
```
neongreen: I feel that this should be a standalone function (as it's useful for both `removeCreds` and `loginUserCreds`).

src/Guide/State.hs
```diff
@@ -860,6 +875,38 @@ logoutUserGlobally key = do
...
+                      = False
+                    | otherwise = True
+
+-- | Get user from creds
+loginUserCreds :: Creds -> Acid.Query GlobalState (Maybe User)
+loginUserCreds userCreds = do
+  matches <- filter (\(_, cs) -> any matchCred cs) . M.toList <$> view creds
+  case matches of
+    [(key, _)] -> view (users . at key)
+    _ -> return Nothing
```
neongreen: Am I right that if there is more than one match, it's a bug? If yes, we should `error` out so that later we'll see in server logs that there was a bug.

src/Guide/Types/AccessToken.hs
```diff
@@ -0,0 +1,36 @@
+{- |
+Module      :  Guide.Config.OAuth2.AccessToken
+Description :  OAuth2 Access Token
+Copyright   :  (c) Aaron Friel
+License     :  BSD-3
+
+Maintainer  :  Aaron Friel <mayreply@aaronfriel.com>
+Stability   :  unstable | experimental | provisional | stable | frozen
```
neongreen: ^

src/Guide/Types/Creds.hs
```diff
@@ -0,0 +1,92 @@
+{- |
+Module      :  Guide.Types.Creds
+Description :  Credentials for external providers (e.g.: GitHub authentication)
+Copyright   :  (c) Aaron Friel
+License     :  BSD-3
+
+Maintainer  :  Aaron Friel <mayreply@aaronfriel.com>
+Stability   :  unstable | experimental | provisional | stable | frozen
```
neongreen: ^

src/Guide/Types/Creds.hs
```diff
@@ -0,0 +1,92 @@
...
+
+-- acid-state
+import Data.SafeCopy hiding (kind)
+import Data.SafeCopy.Migrate
+--
+import qualified Data.Map as Map
+
+data Creds = Creds {
+  -- | The external provider of the credentials (e.g.: github.com)
+  _credsProvider :: Text,
```
neongreen: Perhaps it makes sense to make it an ADT given that we only support a few providers?

src/Guide/Views/Auth/Profile.hs
```diff
@@ -0,0 +1,16 @@
+{- |
+Module      :  Guide.Views.Auth.Profile
+Description :  User profile view
+Copyright   :  (c) Aaron Friel
+License     :  BSD-3
+
+Maintainer  :  Aaron Friel <mayreply@aaronfriel.com>
+Stability   :  unstable | experimental | provisional | stable | frozen
```
neongreen: ^

neongreen:
@AaronFriel I left some nitpicks

src/Guide/Auth/OAuth2/GitHub.hs
```diff
@@ -0,0 +1,191 @@
...
+Copyright   :  (c) Aaron Friel
+License     :  BSD-3
+
+Maintainer  :  Aaron Friel <mayreply@aaronfriel.com>
+Stability   :  unstable
+Portability :  portable
+
+-}
+
+module Guide.Auth.OAuth2.GitHub where
```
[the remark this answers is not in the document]
AaronFriel: Will do.

guide.cabal
```diff
@@ -38,7 +38,7 @@ executable guide
   build-depends:       base
                      , guide
   ghc-options:         -Wall -fno-warn-unused-do-bind
-                       -threaded "-with-rtsopts=-T -N"
+                      --  -threaded "-with-rtsopts=-T -N"
```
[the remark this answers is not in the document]
AaronFriel: I will uncomment them out. I had trouble using the tool `ghcid` for a fast development loop with `-threaded` but I think I was able to resolve that. 

I was using this to develop the site with automatic reloads on code changes:

```
ghcid -c "stack ghci +RTS -T" --test "main" -W -h 52
```

src/Guide/Auth/OAuth2/GitHub.hs
```diff
@@ -0,0 +1,191 @@
...
+        setupCallback = Spock.getpost callbackRoute $ checkCallbackCsrf $ do
+          code :: Text <- param' "code"
+          csrfValue <- getCsrfToken
+          let oauth' = withCallback csrfValue
+          result <- liftIO $ fetchAccessToken mgr oauth' (T.encodeUtf8 code)
+          case result of
+            Left _ -> errorInvalidToken
+            Right token -> do
+              mbCreds <- liftIO $ fetchGithubProfile mgr token
+              case mbCreds of
```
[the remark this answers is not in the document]
AaronFriel: Yes, precisely. I did not modify the registration flow to allow registering accounts via GitHub, or created a profile page for linking and unlinking accounts.

src/Guide/Auth/OAuth2/GitHub.hs
```diff
@@ -0,0 +1,191 @@
...
+                        Nothing -> errorNoAssociatedUser
+                Nothing -> errorNoExternalUser
+        checkCallbackCsrf successAction = do
+          csrf <- getCsrfToken
+          clientCsrf <- param' csrfTokenParam
+          if clientCsrf == csrf
+          then successAction
+          else errorInvalidCsrf
+        csrfTokenParam = "state"
+        errorInvalidCsrf = abort "Broken/invalid CSRF token (Guide error)"
```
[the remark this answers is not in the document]
AaronFriel: Sure, I'll add more to the front-end and expose these messages above the login or registration page, as necessary.

src/Guide/Auth/OAuth2/GitHub.hs
```diff
@@ -0,0 +1,191 @@
...
+    (Left _err, _) -> return $ Nothing -- throwIO $ InvalidProfileResponse "github" err
+    (_, Left _err) -> return $ Nothing -- throwIO $ InvalidProfileResponse "github" err
+
+toCreds :: GithubUser -> [GithubUserEmail] -> AccessToken -> Creds
+toCreds user userMails token = makeCreds "github" ident extra
+    where
+      ident = T.pack $ show $ githubUserId user
+      extra = Map.fromList $
+
+        Nothing
```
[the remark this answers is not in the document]
AaronFriel: Ah, this was preliminary, I needed to add the fields that may be empty. Good catch.

src/Guide/Config/OAuth2/GitHub.hs
AaronFriel: This is for future extensibility, right now the GitHub config just inherits the base config precisely, but I want it to be clear that we may need to do `toJSON ...` twice, and merge the object hashtables. I'll add a comment. (If we later want to add GitHub-specific configuration, it would go here.)

src/Guide/Config/OAuth2/Google.hs
AaronFriel: Just lazy development, I wanted to push something that worked and clean it up later.

src/Guide/State.hs
AaronFriel: Replaced with an `Eq` instance.

src/Guide/Config/OAuth2/GitHub.hs
AaronFriel: I just went ahead and implemented the extensibility with a `scopes` parameter, so you can see why this is there.

AaronFriel:
Except for the front-end work, I believe I've addressed all your critiques, none of which I would call nitpicks.

src/Guide/Types/Creds.hs
AaronFriel: Some providers (e.g.: GitLab, Azure, Google GSuite etc.) might have multiple instances, disambiguated by the "name" parameter on the config.

PR #211, opened 2017-10-08
Force use of `UTF-8`

sectore:
to avoid issue of ` hGetContents: invalid argument (invalid byte sequence)`, which seems to be on Ubuntu only (see discussion at Slack https://aelve.slack.com/archives/C3UQX6F45/p1507399615000050).

_Note:_ `Hakyll` recommands to do the same thing in its `faq` "HGETCONTENTS: INVALID ARGUMENT” OR “COMMITBUFFER: INVALID ARGUMENT" https://jaspervdj.be/hakyll/tutorials/faq.html#hgetcontents-invalid-argument-or-commitbuffer-invalid-argument  _I know we are not using Hakyll :)_

### DIFF
--- a/src/Guide/Main.hs
+++ b/src/Guide/Main.hs
@@ -53,6 +53,7 @@ import Data.Serialize.Get as Cereal
 -- IO
 import System.IO
 import qualified SlaveThread as Slave
+import GHC.IO.Encoding (setLocaleEncoding, utf8)
 -- Catching signals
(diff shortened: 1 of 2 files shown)


neongreen:
Thanks! (Btw, in the future please make comments a bit more detailed, e.g. it would've been nice to mention that it's a workaround for the `hGetContents: invalid argument (invalid byte sequence)` error)

PR #221, opened 2018-10-02
Expose ToC for notes in Items

vrom911:
I've checked it manually, at least I can see the entry like this in the response:
```
... "toc":
  [ [ { "text": "item-notes-ajh4pkjz-types-and-modules"
      , "html":"Types and modules"
      }
    , []
    ]
  , [ { "text": "item-notes-ajh4pkjz-unsafe-operations"
      , "html": "Unsafe operations"
      }
    , []
    ]
  ],
```

### DIFF
--- a/src/Guide/Api/Types.hs
+++ b/src/Guide/Api/Types.hs
@@ -253,7 +253,7 @@ data CItemFull = CItemFull
   , cifNotes       :: CMarkdown                ? "Notes (Markdown)"
   , cifLink        :: Maybe Url                ? "Link to the official site, if exists"
   , cifKind        :: ItemKind                 ? "Item kind, e.g. library, ..."
-  , cifToc         :: Forest (CMarkdown, Text) ? "Table of Contents"
+  , cifToc         :: Forest CHeading          ? "Table of contents"
   } deriving (Show, Generic)
 
 instance A.ToJSON CItemFull where
@@ -290,9 +290,7 @@ toCItemFull Item{..} = CItemFull
   , cifToc         = H $ map treeToCMD (markdownTreeMdTOC _itemNotes)
   }
   where
-    treeToCMD = fmap
-        (\(Heading mdIn slug) -> (toCMarkdown mdIn, slug)
-        )
+    treeToCMD = fmap toCHeading
 
 -- | Client type of 'Trait'
 data CTrait = CTrait
@@ -342,6 +340,20 @@ instance ToCMarkdown MarkdownTree where
     { text = H $ md^.mdSource
     , html = H $ toText . renderText $ toHtml md
     }
+
+data CHeading = CHeading
+  { chContent :: CMarkdown    ? "Rendered heading"
+  , chSlug    :: Text         ? "In-page anchor for linking"
+  } deriving (Show, Generic)
+
+instance A.ToJSON CHeading
+instance ToSchema CHeading
+
+toCHeading :: Heading -> CHeading
+toCHeading h = CHeading
+  { chContent = H $ toCMarkdown $ headingMd h
+  , chSlug    = H $ headingSlug h
+  }
 
 ----------------------------------------------------------------------------
 -- Search client types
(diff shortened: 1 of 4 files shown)


src/Guide/Api/Types.hs
```diff
@@ -1,10 +1,10 @@
+{-# LANGUAGE DataKinds             #-}
 {-# LANGUAGE DeriveGeneric         #-}
-{-# LANGUAGE TypeFamilies          #-}
 {-# LANGUAGE FlexibleInstances     #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE DataKinds             #-}
-{-# LANGUAGE TypeOperators         #-}
 {-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE TypeOperators         #-}
```
ChShersh: We probably should move some of these extensions to the default ones

src/Guide/Api/Types.hs
```diff
@@ -36,18 +36,26 @@ module Guide.Api.Types
...
+import Guide.Markdown (MarkdownBlock, MarkdownInline, MarkdownTree (markdownTreeMdTOC), mdHtml,
+                       mdSource)
+import Guide.Search (SearchResult (..))
+import Guide.Types.Core (Category (..), CategoryStatus, Item (..), ItemKind, Trait (..), content,
+                         ecosystem, uid)
+import Guide.Utils (Uid (..), Url)
+
+import qualified Data.Aeson as A
+import qualified Data.Swagger as S
+import qualified Guide.Types.Core as G
```
ChShersh: ASG

neongreen:
If you're going to use stylish-haskell then maybe let's apply the config to the whole project and then rebase this PR on top?

neongreen:
(I haven't looked at the code yet, only at the sample response above)

src/Guide/Api/Types.hs
neongreen: Yep.

vrom911:
@ChShersh , @neongreen I rebased/rewrote the previous version.

src/Guide/Api/Types.hs
```diff
@@ -284,7 +287,14 @@ toCItemFull Item{..} = CItemFull
   , cifNotes       = H $ toCMarkdown _itemNotes
   , cifLink        = H $ _itemLink
   , cifKind        = H $ _itemKind
+  , cifToc         = H $ map treeToCMD (markdownTreeMdTOC _itemNotes)
   }
+  where
+    treeToCMD = fmap
+        (\(nodes, source) -> CMarkdown
```
neongreen: It's not `(nodes, source)`. It's `(nodes, slug)` (where "slug" is the part after `#` in the link). I guess it's be nice to put that into an ADT, because currently the definition of `MarkdownTree` doesn't make it clear at all and there are no comments either.

To get the source as well, you should change `[Node]` to `[MarkdownInline]` in the definition of `MarkdownTree`.

neongreen: > I guess it's be nice to put that into an ADT

This will also give us a nicer schema for free. (Useful for the frontend)

neongreen: NB: the slug is needed on the frontend as well, of course, because otherwise it can't generate a link.

ChShersh: > To get the source as well, you should change [Node] to [MarkdownInline] in the definition of MarkdownTree

Doing this probably results in DB migration which is not a pleasant thing to do. I think it should be okay for now to do some calculations on the fly and change type later when we will switch Database.

neongreen: Migrating `MarkdownTree` is not complicated (because you don't need to migrate any types that contain `MarkdownTree`, only the type itself), but sure, we can also do calculations on the fly.

The problem is that you can't get the source from `Node` on the fly, not faithfully anyway. Or at least I don't see how this could be done.

ChShersh: @neongreen Could you, please, provide more details on the expected type of the `MarkdownTree` fields and where is the place to perform migration? So we can create smaller subtasks and paralleliese iisues

PR #223, opened 2018-10-06
[WIP] Add custom monad for servant handlers

ChShersh:
The PR is not finished, I need to implement couple more functions and replace current `Handler` with `GuideM` monad.

This monad will be used instead of `Handler` in servant API handlers. It's a first part towards adding `addEdit` to all handlers. This monad will carry `Config` to allow handlers access `Config`.

### DIFF
--- a/guide.cabal
+++ b/guide.cabal
@@ -45,6 +45,8 @@ executable guide
 library
   exposed-modules:
     Guide.App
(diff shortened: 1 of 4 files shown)


willbasky:
done [#245](https://github.com/aelve/guide/pull/245)

PR #226, opened 2018-10-17
Add createTrait

willbasky:


willbasky:
I am not sure about function name. It either create and edit.

src/Guide/Api/Methods.hs
```diff
@@ -104,6 +104,12 @@ deleteItem db itemId = uncache db (CacheItem itemId) $ do
 
 -- TODO: move a trait
 
+-- | Create a trait (pro/con).
```
neongreen, line 107: Nope, `SetTraitContent` only changes an existing trait (e.g. when you click the little pencil icon on a trait). It doesn't create a trait.

neongreen:
```suggestion
-- | Update the text of a trait (pro/con).
```

src/Guide/Api/Methods.hs
```diff
@@ -104,6 +104,12 @@ deleteItem db itemId = uncache db (CacheItem itemId) $ do
...
+setTrait :: DB -> Uid Item -> Uid Trait -> Text -> Handler (Uid Trait)
+setTrait db itemId traitId text = do
+    (_edit, _newItem) <- dbUpdate db (SetTraitContent itemId traitId text)
```
neongreen:
```suggestion
    (_edit, _newTrait) <- dbUpdate db (SetTraitContent itemId traitId text)
    -- TODO: addEdit edit
```

src/Guide/Api/Methods.hs
```diff
@@ -104,6 +104,12 @@ deleteItem db itemId = uncache db (CacheItem itemId) $ do
 
 -- TODO: move a trait
 
+-- | Create a trait (pro/con).
+setTrait :: DB -> Uid Item -> Uid Trait -> Text -> Handler (Uid Trait)
```
neongreen: You can just do `Handler NoContent` – the UID didn't change so there's no need to return it.

src/Guide/Api/Types.hs
```diff
@@ -134,7 +134,18 @@ data ItemSite route = ItemSite
...
-  { _deleteTrait :: route :-
+  { _setTrait :: route :-
+      Summary "Create a new trait in the given item"
+      :> Description "Returns the ID of the created trait."
+      :> ErrorResponse 400 "'text' not provided"
+      :> "item"
+      :> Capture "item" (Uid Item)
+      :> "trait"
+      :> Capture "id" (Uid Trait)
+      :> QueryParam' '[Required, Strict] "text" Text
```
neongreen: It's better to use a `Put` there with the text being contained in the request body (otherwise it's going to end up in the URL; it's not particularly bad, but putting it into the body is a more standard practice)

neongreen:
> It either create and edit.

You can't do creation and editing in one handler. For creation you _can't_ supply the trait ID, for instance, because it has to be generated by the server. For editing, on the other hand, you need the trait ID.

neongreen:
Btw, please move the task in Notion from "Next up" to "In progress".

src/Guide/Api/Methods.hs
```diff
@@ -104,6 +104,23 @@ deleteItem db itemId = uncache db (CacheItem itemId) $ do
...
+    traitId <- randomShortUid
+    (_edit, _newTrait) <- case traitType of
+        Con -> dbUpdate db (AddCon itemId traitId text)
+        Pro -> dbUpdate db (AddPro itemId traitId text)
+    invalidateCache' db (CacheItemTraits itemId)
+    pure traitId
+
+-- | Update the text of a trait (pro/con).
+setTrait :: DB -> Uid Item -> Uid Trait -> Text -> Handler NoContent
+setTrait db itemId traitId text = do
```
neongreen: For modification use `uncache`, not `invalidateCache'`

src/Guide/Api/Methods.hs
```diff
@@ -104,6 +104,23 @@ deleteItem db itemId = uncache db (CacheItem itemId) $ do
...
 -- TODO: move a trait
 
+-- | Create a trait (pro/con).
+createTrait :: DB -> Uid Item -> TraitType -> Text -> Handler (Uid Trait)
+createTrait db itemId traitType text = do
+    traitId <- randomShortUid
+    (_edit, _newTrait) <- case traitType of
+        Con -> dbUpdate db (AddCon itemId traitId text)
+        Pro -> dbUpdate db (AddPro itemId traitId text)
+    invalidateCache' db (CacheItemTraits itemId)
```
neongreen, line 114: Let's have a TODO here for `addEdit edit`, so that we won't forget to do it later.

src/Guide/Api/Types.hs
```diff
@@ -134,7 +135,28 @@ data ItemSite route = ItemSite
...
-  { _deleteTrait :: route :-
+  { _createTrait :: route :-
+      Summary "Create a new trait in the given item"
+      :> Description "Returns the ID of the created trait."
+      :> ErrorResponse 400 "'text' not provided"
+      :> "item"
+      :> Capture "item" (Uid Item)
+      :> "trait"
+      :> Capture "type" TraitType
+      :> QueryParam' '[Required, Strict] "text" Text
```
neongreen: This `QueryParam'` should also become a `ReqBody`

src/Guide/Api/Types.hs
```diff
@@ -134,7 +135,28 @@ data ItemSite route = ItemSite
...
+      :> Post '[JSON] (Uid Trait)
+
+  , _setTrait :: route :-
+      Summary "Update a trait in the given item"
+      :> Description "Returns nothing"
```
neongreen: Well, this description is not useful because it would be very clear from the schema, so it's better to remove it.

src/Guide/Api/Types.hs
```diff
@@ -158,6 +180,27 @@ data SearchSite route = SearchSite
 
 type Api = ToServant Site AsApi
 
+-- | Trait type (Pro/Con) and instancies.
```
neongreen: "instances"

src/Guide/Api/Types.hs
```diff
@@ -158,6 +180,27 @@ data SearchSite route = SearchSite
...
+instance ToSchema TraitType where
+    declareNamedSchema = genericDeclareNamedSchema schemaOptions
+
+instance ToParamSchema TraitType where
+    toParamSchema _ = mempty
+        & S.type_ .~ SwaggerString
+        & S.format ?~ "Trait type"
+
+instance ToHttpApiData TraitType where
+    toUrlPiece = toText . show
```
neongreen: Can you make it lowercase for consistency? `/item/:id/trait/Pro` works, of course, but looks weird.

src/Guide/Api/Types.hs
```diff
@@ -158,6 +180,27 @@ data SearchSite route = SearchSite
...
+
+instance FromHttpApiData TraitType where
+    parseUrlPiece t = case t of
+        "Pro" -> Right Pro
```
neongreen: And here.

neongreen:
Almost there! 🙂

neongreen (approved):

PR #227, opened 2018-10-21
Add setCategoryNote and setCategoryInfo and setItemInfo

willbasky:
I have found a bug. When `setCategoryNote` used from swagger, nothing happened in the description after refresh. However clicking on 'edit' it and then 'save' starts diff view up and new changes appears there.

### DIFF
--- a/src/Guide/Api/Methods.hs
+++ b/src/Guide/Api/Methods.hs
@@ -13,11 +13,13 @@ import Data.Text (Text)
 import Servant
 
 import Guide.Api.Types
+import Guide.Api.Utils
 import Guide.Cache
 import Guide.State
 import Guide.Types
 import Guide.Utils
 
+import qualified Data.Set as S
 import qualified Data.Text as T
 import qualified Guide.Search as Search
 
@@ -63,6 +65,20 @@ setCategoryNotes :: DB -> Uid Category -> Text -> Handler NoContent
 setCategoryNotes db catId note = uncache db (CacheCategoryNotes catId) $ do
     (_edit, _newCategory) <- dbUpdate db (SetCategoryNotes catId note)
     pure NoContent
+
+-- | Edit category`s info (title, group, status, sections (pro/con, ecosystem, note))
+setCategoryInfo :: DB -> Uid Category -> CCategoryInfoEdit -> Handler NoContent
+setCategoryInfo db catId CCategoryInfoEdit {..} =
+    uncache db (CacheCategoryInfo catId) $ do
+        _ <- dbUpdate db $ (\(H cTitle) -> SetCategoryTitle catId cTitle) ccieTitle
+        _ <- dbUpdate db $ (\(H cGroup) -> SetCategoryGroup catId cGroup) ccieGroup
+        _ <- dbUpdate db $ (\(H cStatus) -> SetCategoryStatus catId cStatus) ccieStatus
+        oldEnabledSections <- view enabledSections <$> dbQuery db (GetCategory catId)
+        let newEnabledSections = (\(H sections) -> sections) ccieSections
+        _ <- dbUpdate db $ ChangeCategoryEnabledSections catId
+            (newEnabledSections S.\\ oldEnabledSections)
+            (oldEnabledSections S.\\ newEnabledSections)
+        pure NoContent
 
 -- | Delete a category.
 deleteCategory :: DB -> Uid Category -> Handler NoContent
--- a/src/Guide/Api/Methods.hs
+++ b/src/Guide/Api/Methods.hs
@@ -133,7 +133,7 @@ createItem db catId name' = do
 setItemInfo :: DB -> Uid Item -> CItemInfo -> Handler NoContent
 setItemInfo db itemId CItemInfo{..} = uncache db (CacheItemInfo itemId) $ do
   dbQuery db (GetItemMaybe itemId) >>= \case
-    Nothing -> throwError (err400 {errBody = "Item not found"})
+    Nothing -> throwError (err404 {errBody = "Item not found"})
     Just _ -> do
       -- TODO diff and merge
       _ <- dbUpdate db $ SetItemName itemId $ unH ciiName
--- a/src/Guide/Api/Methods.hs
+++ b/src/Guide/Api/Methods.hs
@@ -82,7 +82,7 @@ setCategoryInfo db catId CCategoryInfoEdit{..} = uncache db (CacheCategoryInfo c
       _ <- dbUpdate db $ SetCategoryTitle catId $ unH ccieTitle
       _ <- dbUpdate db $ SetCategoryGroup catId $ unH ccieGroup
       _ <- dbUpdate db $ SetCategoryStatus catId $ unH ccieStatus
-      oldEnabledSections <- view enabledSections <$> pure category
+      let oldEnabledSections = category ^. enabledSections
       let newEnabledSections = unH ccieSections
       _ <- dbUpdate db $ ChangeCategoryEnabledSections catId
           (newEnabledSections S.\\ oldEnabledSections)
--- a/src/Guide/Api/Methods.hs
+++ b/src/Guide/Api/Methods.hs
@@ -58,6 +58,12 @@ createCategory db title' group' = do
       -- TODO addEdit edit
       return catId
 
+-- | Edit categoty's note.
+setCategoryNote :: DB -> Uid Category -> Text -> Handler NoContent
+setCategoryNote db catId note = uncache db (CacheCategory catId) $ do
+    (_edit, _newCategory) <- dbUpdate db (SetCategoryNotes catId note)
+    pure NoContent
+
 -- | Delete a category.
 deleteCategory :: DB -> Uid Category -> Handler NoContent
 deleteCategory db catId = uncache db (CacheCategory catId) $ do
--- a/src/Guide/Api/Methods.hs
+++ b/src/Guide/Api/Methods.hs
@@ -94,7 +94,7 @@ setCategoryInfo db catId CCategoryInfoEdit{..} = uncache db (CacheCategoryInfo c
 deleteCategory :: DB -> Uid Category -> Handler NoContent
 deleteCategory db catId = uncache db (CacheCategory catId) $ do
   dbQuery db (GetCategoryMaybe catId) >>= \case
-    Nothing -> throwError (err400 {errBody = "Category not provided"})
+    Nothing -> throwError (err404 {errBody = "Category not found"})
     Just _ -> do
       _mbEdit <- dbUpdate db (DeleteCategory catId)
       pure NoContent
--- a/src/Guide/Api/Methods.hs
+++ b/src/Guide/Api/Methods.hs
@@ -158,7 +158,7 @@ deleteItem db itemId = uncache db (CacheItem itemId) $ do
 -- | Create a trait (pro/con).
 createTrait :: DB -> Uid Item -> TraitType -> Text -> Handler (Uid Trait)
 createTrait db itemId traitType text = do
-  when (T.null text) $ do throwError (err400 {errBody = "Name not provided"})
+  when (T.null text) $ throwError (err400 {errBody = "Trait text not provided"})
   traitId <- randomShortUid
   (_edit, _newTrait) <- case traitType of
     Con -> dbUpdate db (AddCon itemId traitId text)
--- a/src/Guide/Api/Methods.hs
+++ b/src/Guide/Api/Methods.hs
@@ -111,7 +111,7 @@ deleteCategory db catId = uncache db (CacheCategory catId) $ do
 createItem :: DB -> Uid Category -> Text -> Handler (Uid Item)
 createItem db catId name' = do
   dbQuery db (GetCategoryMaybe catId) >>= \case
-    Nothing -> throwError (err400 {errBody = "Category not found"})
+    Nothing -> throwError (err404 {errBody = "Category not found"})
     Just _ -> do
       if T.null name' then throwError (err400 {errBody = "Name not provided"})
       else do
--- a/src/Guide/Api/Methods.hs
+++ b/src/Guide/Api/Methods.hs
@@ -87,6 +87,7 @@ setCategoryInfo db catId CCategoryInfoEdit{..} = uncache db (CacheCategoryInfo c
       _ <- dbUpdate db $ ChangeCategoryEnabledSections catId
           (newEnabledSections S.\\ oldEnabledSections)
           (oldEnabledSections S.\\ newEnabledSections)
+      -- TODO record edits
       pure NoContent
 
 -- | Delete a category.
--- a/src/Guide/Api/Methods.hs
+++ b/src/Guide/Api/Methods.hs
@@ -74,7 +74,7 @@ setCategoryNotes db catId note = uncache db (CacheCategoryNotes catId) $ do
 
 -- | Edit category's info (title, group, status, sections (pro/con, ecosystem, note)).
 setCategoryInfo :: DB -> Uid Category -> CCategoryInfoEdit -> Handler NoContent
-setCategoryInfo db catId CCategoryInfoEdit {..} = uncache db (CacheCategoryInfo catId) $ do
+setCategoryInfo db catId CCategoryInfoEdit{..} = uncache db (CacheCategoryInfo catId) $ do
     dbQuery db (GetCategoryMaybe catId) >>= \case
         Nothing -> throwError (err404 {errBody = "Category not found"})
         Just _ -> do
@@ -127,6 +127,12 @@ createItem db catId name' = do
                 pure itemId
 
 -- TODO: move an item
+
+-- | Set item`s info
+setItemInfo :: DB -> Uid Item -> CItemInfo -> Handler NoContent
+setItemInfo db itemId CItemInfo{..} = uncache db (CacheItemInfo itemId) $ do
+    dbQuery db (GetItem itemId) >>= \case
+        Nothing -> throwError (err404 {errBody = "Category not found"})
 
 -- | Delete an item.
 deleteItem :: DB -> Uid Item -> Handler NoContent
--- a/src/Guide/Api/Methods.hs
+++ b/src/Guide/Api/Methods.hs
@@ -46,46 +46,58 @@ getCategory db catId =
 -- category with this title exists already).
 createCategory :: DB -> Text -> Text -> Handler (Uid Category)
 createCategory db title' group' = do
-  -- If the category exists already, don't create it
-  cats <- view categories <$> dbQuery db GetGlobalState
-  let isDuplicate cat = T.toCaseFold (cat^.title) == T.toCaseFold title'
-                     && T.toCaseFold (cat^.group_) == T.toCaseFold group'
-  case find isDuplicate cats of
-    Just c  -> return (c^.uid)
-    Nothing -> do
-      catId <- randomShortUid
-      time <- liftIO getCurrentTime
-      (_edit, _newCategory) <- dbUpdate db (AddCategory catId title' group' time)
-      invalidateCache' db (CacheCategory catId)
-      -- TODO addEdit edit
-      return catId
+    if T.null title' then throwError (err404 {errBody = "Title is empty"})
+    else do
+        -- If the category exists already, don't create it
+        cats <- view categories <$> dbQuery db GetGlobalState
+        let isDuplicate cat = T.toCaseFold (cat^.title) == T.toCaseFold title'
+                           && T.toCaseFold (cat^.group_) == T.toCaseFold group'
+        case find isDuplicate cats of
+          Just c  -> return (c^.uid)
+          Nothing -> do
+            catId <- randomShortUid
+            time <- liftIO getCurrentTime
+            (_edit, _newCategory) <- dbUpdate db (AddCategory catId title' group' time)
+            invalidateCache' db (CacheCategory catId)
+            -- TODO addEdit edit
+            return catId
 
 -- | Edit categoty's note.
 setCategoryNotes :: DB -> Uid Category -> Text -> Handler NoContent
 setCategoryNotes db catId note = uncache db (CacheCategoryNotes catId) $ do
-    (_edit, _newCategory) <- dbUpdate db (SetCategoryNotes catId note)
-    pure NoContent
+    dbQuery db (GetCategoryMaybe catId) >>= \case
+        Nothing -> throwError (err404 {errBody = "Category not found"})
+        Just _ -> do
+            (_edit, _newCategory) <- dbUpdate db (SetCategoryNotes catId note)
+    -- TODO diff and merge
+            pure NoContent
 
--- | Edit category`s info (title, group, status, sections (pro/con, ecosystem, note))
+-- | Edit category's info (title, group, status, sections (pro/con, ecosystem, note)).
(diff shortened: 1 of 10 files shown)


src/Guide/Api/Methods.hs
```diff
@@ -58,6 +58,12 @@ createCategory db title' group' = do
       -- TODO addEdit edit
       return catId
 
+-- | Edit categoty's note.
+setCategoryNote :: DB -> Uid Category -> Text -> Handler NoContent
```
neongreen: Notes, plural

src/Guide/Api/Types.hs
```diff
@@ -106,6 +106,14 @@ data CategorySite route = CategorySite
            "group" Text
       :> Post '[JSON] (Uid Category)
 
+  , _setCategoryNote :: route :-
+      Summary "Edit category`s note"
```
neongreen: That's some weird apostrophe

src/Guide/Api/Types.hs
```diff
@@ -106,6 +106,14 @@ data CategorySite route = CategorySite
...
+      :> ErrorResponse 404 "Category not found"
+      :> "category"
+      :> Capture "id" (Uid Category)
```
neongreen, line 116: The route should somehow mention that we're editing notes

neongreen (changes requested):

src/Guide/Api/Methods.hs
```diff
@@ -58,6 +60,26 @@ createCategory db title' group' = do
       -- TODO addEdit edit
       return catId
 
+-- | Edit categoty's note.
+setCategoryNotes :: DB -> Uid Category -> Text -> Handler NoContent
+setCategoryNotes db catId note = uncache db (CacheCategoryNotes catId) $ do
+    (_edit, _newCategory) <- dbUpdate db (SetCategoryNotes catId note)
+    pure NoContent
+
+-- | Edit category`s info (title, group, status, sections (pro/con, ecosystem, note))
```
neongreen: `category's`

src/Guide/Api/Methods.hs
```diff
@@ -58,6 +60,26 @@ createCategory db title' group' = do
...
+setCategoryInfo :: DB -> Uid Category -> CCategoryInfoEdit -> Handler NoContent
+setCategoryInfo db catId CCategoryInfoEdit {..} =
+    uncache db (CacheCategoryInfo catId) $ do
+        _ <- dbUpdate db $ (\(H cTitle) -> SetCategoryTitle catId cTitle) ccieTitle
```
neongreen: I would recommend adding something like `unH` and then you can write these in a nicer way:

```haskell
_ <- dbUpdate db $ SetCategoryTitle catId (unH ccieTitle)
```

src/Guide/Api/Types.hs
```diff
@@ -106,6 +109,27 @@ data CategorySite route = CategorySite
            "group" Text
       :> Post '[JSON] (Uid Category)
 
+  , _setCategoryNotes :: route :-
+      Summary "Edit category's note"
```
neongreen: "notes"

src/Guide/Api/Types.hs
```diff
@@ -106,6 +109,27 @@ data CategorySite route = CategorySite
...
+      :> ErrorResponse 404 "Category not found"
+      :> "category"
+      :> Capture "id" (Uid Category)
+      :> "notes"
+      :> ReqBody '[JSON] Text
+      :> Put '[JSON] NoContent
+
+  , _setCategoryInfo :: route :-
+      Summary "Set category's fields"
+      :> Description "Set category fields:\
```
neongreen: This description is not needed because the schema for `CCategoryInfoEdit` will already mention that.

src/Guide/Api/Types.hs
```diff
@@ -266,6 +294,40 @@ toCCategoryFull Category{..} = CCategoryFull
   , ccfStatus      = H $ _categoryStatus
   }
 
+data CCategoryInfoEdit = CCategoryInfoEdit
+    { ccieTitle    :: Text           ? "Category title"
+    , ccieGroup    :: Text           ? "Category group ('grandcategory')"
+    , ccieStatus   :: CategoryStatus ? "Status (done, in progress, ...)"
+    , ccieSections :: (Set ItemSection) ? "Enable/disable item sections"
```
neongreen: Unnecessary parentheses.

neongreen: Also, "enabled item sections".

src/Guide/Api/Methods.hs
```diff
@@ -58,6 +60,26 @@ createCategory db title' group' = do
       -- TODO addEdit edit
       return catId
 
+-- | Edit categoty's note.
+setCategoryNotes :: DB -> Uid Category -> Text -> Handler NoContent
+setCategoryNotes db catId note = uncache db (CacheCategoryNotes catId) $ do
+    (_edit, _newCategory) <- dbUpdate db (SetCategoryNotes catId note)
```
neongreen: Please add a TODO that we want to handle diffs and merges here.

src/Guide/Api/Methods.hs
```diff
@@ -58,6 +60,26 @@ createCategory db title' group' = do
       -- TODO addEdit edit
       return catId
 
+-- | Edit categoty's note.
+setCategoryNotes :: DB -> Uid Category -> Text -> Handler NoContent
+setCategoryNotes db catId note = uncache db (CacheCategoryNotes catId) $ do
```
neongreen, line 67: The endpoint's description says that it returns 404 when the category is not found, but it does not. Please add checks and `throwError (err404 {errBody = "Category not found"})` to both endpoints (and to all other endpoints where the description says 404 may be returned)

src/Guide/Api/Types.hs
```diff
@@ -19,10 +19,13 @@ module Guide.Api.Types
   , SearchSite(..)
   , Site(..)
   , TraitSite(..)
+
+  -- Type for backend
```
neongreen: `TraitType` is not a "type for backend", it is more-or-less exposed to the frontend in URLs. It's hard to write a good description for this type, but a misleading description is worse than none, so you can just write:

```
-- * Other types
```

Note that you need an `*` there if you want it to be rendered as a separate section in Haddocks.

src/Guide/Api/Types.hs
```diff
@@ -266,6 +294,40 @@ toCCategoryFull Category{..} = CCategoryFull
...
+    , ccieStatus   :: CategoryStatus ? "Status (done, in progress, ...)"
+    , ccieSections :: (Set ItemSection) ? "Enable/disable item sections"
+    }
+    deriving (Show, Generic)
+
+instance A.ToJSON CCategoryInfoEdit where
+  toJSON = A.genericToJSON jsonOptions
+
+instance A.FromJSON CCategoryInfoEdit where
+  parseJSON = A.withObject "CCategoryInfoEdit" $ \o -> do
```
neongreen: Did `A.genericParseJSON jsonOptions` not work?

src/Guide/Types/Core.hs
```diff
@@ -158,6 +158,13 @@ deriveSafeCopySimple 0 'base ''ItemSection
 instance A.ToJSON ItemSection where
   toJSON = A.genericToJSON A.defaultOptions
 
+instance A.FromJSON ItemSection where
+  parseJSON v = case v of
+    "ItemProsConsSection" -> pure ItemProsConsSection
```
neongreen: Again, did `A.genericParseJSON A.defaultOptions` not work?

src/Guide/Types/Core.hs
```diff
@@ -206,6 +213,13 @@ deriveSafeCopySimple 2 'extension ''CategoryStatus
 instance A.ToJSON CategoryStatus where
   toJSON = A.genericToJSON A.defaultOptions
 
+instance A.FromJSON CategoryStatus where
```
neongreen, line 221: Same.

src/Guide/Api/Types.hs
```diff
@@ -266,6 +294,40 @@ toCCategoryFull Category{..} = CCategoryFull
...
+    ccieTitle    <- o A..: "title"
+    ccieGroup    <- o A..: "group"
+    ccieStatus   <- o A..: "status"
+    ccieSections <- o A..: "sections"
+    pure CCategoryInfoEdit{..}
+
+instance ToSchema CCategoryInfoEdit where
+  declareNamedSchema = genericDeclareNamedSchema schemaOptions
+
+-- | Factory to create a 'CCategoryInfoEdit' from a 'Category'
```
neongreen: Where is this function used?

src/Guide/Api/Utils.hs
```diff
@@ -48,6 +48,8 @@ newtype (?) (field :: *) (help :: Symbol) = H field
 instance ToJSON field => ToJSON (field ? help) where
   toJSON (H a) = toJSON a
 
+instance FromJSON field => FromJSON (field ? help)
```
neongreen: This instance should be written explicitly (like `ToJSON`), because otherwise it uses a generic instance and AFAIK wraps the value into something.

Generally, whenever you have two symmetric things (in this case ToJSON and FromJSON) and you use different approaches for them, it might be a sign that something is wrong here. And even if nothing is wrong, being consistent still helps readability.

src/Guide/Api/Methods.hs
```diff
@@ -44,26 +46,58 @@ getCategory db catId =
...
-  case find isDuplicate cats of
-    Just c  -> return (c^.uid)
-    Nothing -> do
-      catId <- randomShortUid
-      time <- liftIO getCurrentTime
-      (_edit, _newCategory) <- dbUpdate db (AddCategory catId title' group' time)
-      invalidateCache' db (CacheCategory catId)
-      -- TODO addEdit edit
-      return catId
+    if T.null title' then throwError (err404 {errBody = "Title is empty"})
```
neongreen: No, 404 is "not found". You want to use 400 here.

Here are some common error codes:

* 400 bad request
* 401 not authenticated (credentials needed but they weren't supplied; or credentials are invalid)
* 403 forbidden (you've logged in successfully but you, the user, are not allowed to access some resource)
* 404 resource not found
* 409 conflict (e.g. you tried to create a category with a title that exists already)
* 500 server error (e.g. a bug or an uncaught exception)

src/Guide/Api/Methods.hs
```diff
@@ -44,26 +46,58 @@ getCategory db catId =
 -- category with this title exists already).
 createCategory :: DB -> Text -> Text -> Handler (Uid Category)
 createCategory db title' group' = do
-  -- If the category exists already, don't create it
```
neongreen: If you want to switch to 4-spaces indentation, it's fine, but it should be done in a separate PR. Otherwise you're just breaking consistency in the code.

src/Guide/Api/Methods.hs
neongreen: Also, you don't need `if then else` here, you can just do:

```haskell
when (T.null title') $
  throwError ...
```

Errors have to be thrown fairly often in handlers; if you use `if` all the time, all your code will go to the right pretty fast.

src/Guide/Api/Methods.hs
```diff
@@ -44,26 +46,58 @@ getCategory db catId =
...
+setCategoryInfo :: DB -> Uid Category -> CCategoryInfoEdit -> Handler NoContent
+setCategoryInfo db catId CCategoryInfoEdit{..} = uncache db (CacheCategoryInfo catId) $ do
+    dbQuery db (GetCategoryMaybe catId) >>= \case
+        Nothing -> throwError (err404 {errBody = "Category not found"})
+        Just _ -> do
+            -- TODO diff and merge
+            _ <- dbUpdate db $ SetCategoryTitle catId $ unH ccieTitle
+            _ <- dbUpdate db $ SetCategoryGroup catId $ unH ccieGroup
+            _ <- dbUpdate db $ SetCategoryStatus catId $ unH ccieStatus
+            oldEnabledSections <- view enabledSections <$> dbQuery db (GetCategory catId)
```
neongreen: You don't need to query anything here – the category is already available because you're in the `Just` branch.

src/Guide/Api/Methods.hs
```diff
@@ -75,22 +109,40 @@ deleteCategory db catId = uncache db (CacheCategory catId) $ do
...
+                    kind' = if looksLikeLibrary then Library (Just name') else Other
+                time <- liftIO getCurrentTime
+                (_edit, _newItem) <- dbUpdate db (AddItem catId itemId name' time kind')
+                invalidateCache' db (CacheItem itemId)
+                -- TODO: addEdit edit
+                pure itemId
 
 -- TODO: move an item
 
+-- | Set item`s info
```
neongreen: Apostrophe instead of a backtick

src/Guide/Api/Types.hs
```diff
@@ -124,6 +145,15 @@ data ItemSite route = ItemSite
       :> QueryParam' '[Required, Strict] "name" Text
       :> Post '[JSON] (Uid Item)
 
+  , _setItemInfo :: route :-
+      Summary "Set item's fields"
+      :> ErrorResponse 404 "Item not found"
```
neongreen, line 151: What about 400 when you try to set the name to empty?

src/Guide/Api/Types.hs
```diff
@@ -265,6 +299,26 @@ toCCategoryFull Category{..} = CCategoryFull
   , ccfStatus      = H $ _categoryStatus
   }
 
+data CCategoryInfoEdit = CCategoryInfoEdit
```
neongreen, line 304: Add a `-- |` comment to the type

src/Guide/Api/Utils.hs
```diff
@@ -58,6 +62,10 @@ instance (KnownSymbol help, ToSchema a) => ToSchema (a ? help) where
 instance {-# OVERLAPPING #-} (KnownSymbol help, Selector s, ToSchema c) => GToSchema (S1 s (K1 i (Maybe c ? help))) where
   gdeclareNamedSchema opts _ = fmap unnamed . withFieldSchema opts (Proxy2 :: Proxy2 s (K1 i (Maybe c ? help))) False
 
+-- | Runner for 'field ? help'
```
neongreen: It's not a runner, it unwraps the newtype. "Runner" is usually used with monads.

src/Guide/Types/Core.hs
```diff
@@ -131,6 +132,15 @@ instance A.ToJSON ItemKind where
   toJSON Other = A.object [
     "tag"      A..= ("Other" :: Text) ]
 
+instance A.FromJSON ItemKind where
+  parseJSON ik@(A.Object o) = do
+    iki <- A.parseJSON ik
```
neongreen: This instance loops indefinitely. `parseJSON ik = do iki <- parseJSON ik`

You should extract the tag from the object, then branch on it.

Also, please use `withObject` instead of inspecting the value manually, it gives nicer error messages (you get told what type the instance was parsing when it failed).

src/Guide/Api/Methods.hs
```diff
@@ -44,10 +46,12 @@ getCategory db catId =
 -- category with this title exists already).
 createCategory :: DB -> Text -> Text -> Handler (Uid Category)
 createCategory db title' group' = do
+  when (T.null title') $ do throwError (err400 {errBody = "Title' not provided"})
```
neongreen:
```suggestion
  when (T.null title') $ do throwError (err400 {errBody = "Title not provided"})
```

src/Guide/Api/Methods.hs
```diff
@@ -58,12 +62,43 @@ createCategory db title' group' = do
...
+    Just _ -> do
+      (_edit, _newCategory) <- dbUpdate db (SetCategoryNotes catId note)
+      -- TODO diff and merge
+      pure NoContent
+
+-- | Edit category's info (title, group, status, sections (pro/con, ecosystem, note)).
+setCategoryInfo :: DB -> Uid Category -> CCategoryInfoEdit -> Handler NoContent
+setCategoryInfo db catId CCategoryInfoEdit{..} = uncache db (CacheCategoryInfo catId) $ do
+  dbQuery db (GetCategoryMaybe catId) >>= \case
+    Nothing -> throwError (err400 {errBody = "Category not provided"})
```
neongreen:
```suggestion
    Nothing -> throwError (err404 {errBody = "Category not found"})
```

src/Guide/Api/Methods.hs
```diff
@@ -58,12 +62,43 @@ createCategory db title' group' = do
...
+    Just category -> do
+      -- TODO diff and merge
+      _ <- dbUpdate db $ SetCategoryTitle catId $ unH ccieTitle
+      _ <- dbUpdate db $ SetCategoryGroup catId $ unH ccieGroup
+      _ <- dbUpdate db $ SetCategoryStatus catId $ unH ccieStatus
+      oldEnabledSections <- view enabledSections <$> pure category
```
neongreen:
```suggestion
      let oldEnabledSections = category ^. enabledSections
```

src/Guide/Api/Methods.hs
```diff
@@ -58,12 +62,43 @@ createCategory db title' group' = do
...
+      pure NoContent
+
 -- | Delete a category.
 deleteCategory :: DB -> Uid Category -> Handler NoContent
 deleteCategory db catId = uncache db (CacheCategory catId) $ do
-  _mbEdit <- dbUpdate db (DeleteCategory catId)
-  pure NoContent
-  -- TODO mapM_ addEdit mbEdit
+  dbQuery db (GetCategoryMaybe catId) >>= \case
+    Nothing -> throwError (err400 {errBody = "Category not provided"})
```
neongreen:
```suggestion
    Nothing -> throwError (err404 {errBody = "Category not found"})
```

src/Guide/Api/Methods.hs
```diff
@@ -75,22 +110,38 @@ deleteCategory db catId = uncache db (CacheCategory catId) $ do
...
-  let isAllowedChar c = isAscii c && (isAlphaNum c || c == '-')
-      looksLikeLibrary = T.all isAllowedChar name'
-      kind' = if looksLikeLibrary then Library (Just name') else Other
-  time <- liftIO getCurrentTime
-  (_edit, _newItem) <- dbUpdate db (AddItem catId itemId name' time kind')
-  invalidateCache' db (CacheItem itemId)
-  -- TODO: addEdit edit
-  pure itemId
+  dbQuery db (GetCategoryMaybe catId) >>= \case
+    Nothing -> throwError (err400 {errBody = "Category not found"})
```
neongreen:
```suggestion
    Nothing -> throwError (err404 {errBody = "Category not found"})
```

src/Guide/Api/Methods.hs
```diff
@@ -75,22 +110,38 @@ deleteCategory db catId = uncache db (CacheCategory catId) $ do
...
+        -- TODO: addEdit edit
+        pure itemId
 
 -- TODO: move an item
 
+-- | Set item's info
+setItemInfo :: DB -> Uid Item -> CItemInfo -> Handler NoContent
+setItemInfo db itemId CItemInfo{..} = uncache db (CacheItemInfo itemId) $ do
+  dbQuery db (GetItemMaybe itemId) >>= \case
+    Nothing -> throwError (err400 {errBody = "Item not found"})
```
neongreen:
```suggestion
    Nothing -> throwError (err404 {errBody = "Item not found"})
```

src/Guide/Api/Methods.hs
```diff
@@ -107,19 +158,21 @@ deleteItem db itemId = uncache db (CacheItem itemId) $ do
...
 createTrait :: DB -> Uid Item -> TraitType -> Text -> Handler (Uid Trait)
 createTrait db itemId traitType text = do
-    traitId <- randomShortUid
-    (_edit, _newTrait) <- case traitType of
-        Con -> dbUpdate db (AddCon itemId traitId text)
-        Pro -> dbUpdate db (AddPro itemId traitId text)
-    invalidateCache' db (CacheItemTraits itemId)
--- TODO: mapM_ addEdit mbEdit
-    pure traitId
+  when (T.null text) $ do throwError (err400 {errBody = "Name not provided"})
```
neongreen:
```suggestion
  when (T.null text) $ throwError (err400 {errBody = "Trait text not provided"})
```

src/Guide/Api/Types.hs
```diff
@@ -265,6 +300,27 @@ toCCategoryFull Category{..} = CCategoryFull
...
+instance A.FromJSON CCategoryInfoEdit where
+  parseJSON = A.genericParseJSON jsonOptions
+
+instance ToSchema CCategoryInfoEdit where
+  declareNamedSchema = genericDeclareNamedSchema schemaOptions
+
+instance ToSchema ItemSection where
+  declareNamedSchema = genericDeclareNamedSchema schemaOptions
+
 -- | A lightweight info type about an 'Item'
```
neongreen:
```suggestion
-- | A lightweight info type about an 'Item'.
--
-- When updating it, don't forget to also update 'setItemInfo'.
```

src/Guide/Api/Utils.hs
```diff
@@ -58,6 +62,10 @@ instance (KnownSymbol help, ToSchema a) => ToSchema (a ? help) where
 instance {-# OVERLAPPING #-} (KnownSymbol help, Selector s, ToSchema c) => GToSchema (S1 s (K1 i (Maybe c ? help))) where
   gdeclareNamedSchema opts _ = fmap unnamed . withFieldSchema opts (Proxy2 :: Proxy2 s (K1 i (Maybe c ? help))) False
 
+-- | Unwraper for 'field ? help'
```
neongreen:
```suggestion
-- | Unwrapper for @field '?' help@
```

neongreen: (Funky Haddock syntax; `@ @` is for monospace, `' '` is for linking to Haskell identifiers)

src/Guide/Types/Core.hs
```diff
@@ -131,6 +131,12 @@ instance A.ToJSON ItemKind where
   toJSON Other = A.object [
     "tag"      A..= ("Other" :: Text) ]
 
+instance A.FromJSON ItemKind where
+  parseJSON = A.withObject "ItemKind" $ \o
```
neongreen: This doesn't work.

`<|>` picks the first parser that succeeds. If there's a field called `"contents"`, the first parser will always succeed, and the second parser will never be chosen.

You want this:

```haskell
  parseJSON = A.withObject "ItemKind" $ \o ->
    o A..: "tag" >>= \case
      "Library" -> Library <$> o A..: "contents"
      "Tool"    -> Tool <$> o A..: "contents"
      "Other"   -> pure Other
      tag       -> fail ("unknown tag " ++ show tag)
```

src/Guide/Api/Methods.hs
```diff
@@ -58,12 +62,43 @@ createCategory db title' group' = do
       -- TODO addEdit edit
       return catId
 
+-- | Edit categoty's note.
+setCategoryNotes :: DB -> Uid Category -> Text -> Handler NoContent
+setCategoryNotes db catId note = uncache db (CacheCategoryNotes catId) $ do
+  dbQuery db (GetCategoryMaybe catId) >>= \case
+    Nothing -> throwError (err400 {errBody = "Category not provided"})
```
neongreen:
```suggestion
    Nothing -> throwError (err404 {errBody = "Category not found"})
```

neongreen (approved):

PR #241, opened 2018-11-13
Add diff and merge to 'setCategoryNotes' and 'setTrait'

willbasky:


src/Guide/Api/Methods.hs
```diff
@@ -65,10 +67,16 @@ setCategoryNotes :: DB -> Uid Category -> Text -> Handler NoContent
 setCategoryNotes db catId note =
   dbQuery db (GetCategoryMaybe catId) >>= \case
     Nothing -> throwError (err404 {errBody = "Category not found"})
-    Just _ -> do
-      (_edit, _newCategory) <- dbUpdate db (SetCategoryNotes catId note)
-      -- TODO diff and merge
-      pure NoContent
+    Just original -> do
+      let fromMarkDown = markdownBlockMdSource . _categoryNotes
```
neongreen: It's "Markdown", not "MarkDown". https://en.wikipedia.org/wiki/Markdown

src/Guide/Api/Methods.hs
```diff
@@ -65,10 +67,16 @@ setCategoryNotes :: DB -> Uid Category -> Text -> Handler NoContent
...
+      modified <- dbQuery db (GetCategory catId)
+      if fromMarkDown original /= fromMarkDown modified then do
+        let merged = merge (fromMarkDown original) note (fromMarkDown modified)
+        (_edit, _newCategory) <- dbUpdate db (SetCategoryNotes catId merged)
```
neongreen: Nope. We don't simply put merged content into the database, because the merge might be crappy.

The merged version (along with the `original`, `modified` and `note`) should be returned in the response, with status code 409 conflict. You can create a client type `CMergeConflict` for that.

src/Guide/Api/Methods.hs
```diff
@@ -204,4 +224,4 @@ dbUpdate db x = liftIO $ do
 dbQuery :: (MonadIO m, EventState event ~ GlobalState, QueryEvent event)
         => DB -> event -> m (EventResult event)
 dbQuery db x = liftIO $
-  Acid.query db x
+  Acid.query db x
```
neongreen: By convention, text files [should have trailing newlines](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file). In VSCode this is configured with `"files.insertFinalNewline": true`.

src/Guide/Api/Methods.hs
```diff
@@ -166,9 +174,21 @@ createTrait db itemId traitType text = do
...
-  -- TODO diff and merge
-  pure NoContent
+  dbQuery db (GetItemMaybe itemId) >>= \case
+    Nothing -> throwError (err404 {errBody = "Item not found"})
+    Just _ -> do
+      dbQuery db (GetTraitMaybe itemId traitId) >>= \case
+        Nothing -> throwError (err404 {errBody = "Trait not found"})
+        Just original -> do
+          let fromMarkDown = markdownInlineMdSource . _traitContent
+          modified <- dbQuery db (GetTrait itemId traitId)
```
neongreen: `modified` and `original` are going to be the same here.

The server doesn't know what the original trait was, because "original" = "what the user saw when they started editing", and the frontend doesn't notify the server when the user starts editing.

You have to expect that the user will send you both `original` and `modified`. You can create a client type for that: `CTextEdit { original :: Text, modified :: Text }`.

src/Guide/Api/Methods.hs
neongreen: Don't forget to update the other handler, too.

src/Guide/Api/Methods.hs
neongreen: Don't forget to update the other handler, too.

neongreen (changes requested):
.

PR #244, opened 2018-11-21
Category opens in vue.js frontend

Gigiman:


front/client/App.vue
```diff
@@ -3,7 +3,9 @@
     <toolbar></toolbar>
     <v-content>
       <v-container fluid>
-        <router-view></router-view>
+        <keep-alive exclude="article-page">
```
avele, line 6: Why do you need keep-alive for almost all the app, except article-page?

front/client/components/Article.vue
```diff
@@ -8,12 +8,12 @@
           v-for="(value, key) in getCategory"
           :key="key"
         >
-          <a
+          <router-link
             class="article-top-link"
-            href="https://guide.aelve.com/haskell/lenses-sth6l9jl"
+            :to="`${pathname()}`"
```
avele: Why use function but not getter? Also bad name, pathname to where?

front/client/components/Article.vue
```diff
@@ -78,8 +81,11 @@ import AddItemDialog from 'client/components/AddItemDialog.vue'
 export default class ArticleItem extends Vue {
   isDialogOpen: boolean = false
 
-  async asyncData () {
-    await this.$store.dispatch('categoryItem/loadCategoryItem')
+  async asyncData() {
+    const rawUrl = this.$route.params.category
```
avele, line 87: You dont need to access router params, just pass it as props. https://router.vuejs.org/guide/essentials/passing-props.html

front/client/App.vue
Gigiman: there is an issue about cache whick I was not able to solve even with keep-alive. Please take a look

PR #245, opened 2018-11-22
Add Guider

willbasky:


src/Guide/Api/Guider.hs
```diff
@@ -0,0 +1,28 @@
+module Guide.Api.Guider
```
neongreen, line 4: Every module should have a `-- |` description.

src/Guide/Api/Guider.hs
```diff
@@ -0,0 +1,28 @@
...
+       ( Guider
+       , guiderToHandler
+       ) where
+
+import Imports
+
+import Control.Monad.IO.Class (MonadIO)
```
neongreen: This is exported by `Imports`, AFAIK.

src/Guide/Api/Guider.hs
```diff
@@ -0,0 +1,28 @@
...
+  { runGuider :: ReaderT Config IO a
+  } deriving (Functor, Applicative, Monad, MonadIO, MonadReader Config)
+
+
+instance MonadError ServantErr Guider where
+  throwError :: ServantErr -> Guider a
+  throwError = liftIO . throwIO
+
+  catchError :: Guider a -> (ServantErr -> Guider a) -> Guider a
+  catchError = error "There is no implementation for catchError"
```
neongreen: Can you implement `catchError`?

src/Guide/Api/Guider.hs
```diff
@@ -0,0 +1,28 @@
...
+import Imports
+
+import Control.Monad.IO.Class (MonadIO)
+import Servant (Handler)
+import Servant.Server.Internal.ServantErr (ServantErr)
+
+import Guide.Config (Config (..), def)
+
+
+newtype Guider a = Guider
```
neongreen, line 19: All public-facing types and functions should have haddocks.

src/Guide/Api/Server.hs
```diff
@@ -30,38 +30,39 @@ import Guide.Api.Methods
 import Guide.Api.Types
 import Guide.Config (Config (..))
 import Guide.State
+import Guide.Api.Guider (guiderToHandler)
 
 import Data.Acid as Acid
 import qualified Data.ByteString.Char8 as BSC
 
 apiServer :: DB -> Site AsServer
```
neongreen: I see that you've decided to avoid using `hoistServer`. Why?

src/Guide/Api/Guider.hs
```diff
@@ -0,0 +1,28 @@
...
+
+instance MonadError ServantErr Guider where
+  throwError :: ServantErr -> Guider a
+  throwError = liftIO . throwIO
+
+  catchError :: Guider a -> (ServantErr -> Guider a) -> Guider a
+  catchError = error "There is no implementation for catchError"
+
+guiderToHandler :: Guider a -> Handler a
+guiderToHandler guider = liftIO $ runReaderT (runGuider guider) def
```
neongreen: This is incorrect. In a `Handler`, you're supposed to return `ServantErr`ors using `ExceptT` instead of throwing them as exceptions. Therefore, you have to catch them here and return properly.

guide.cabal
```diff
@@ -182,6 +183,10 @@ library
                      , DeriveGeneric
                      , TypeApplications
                      , NoImplicitPrelude
+                     , GeneralizedNewtypeDeriving
```
neongreen: Let's not turn on this extension for all modules – it implicitly changes the behavior of `deriving` for newtypes. Other extensions are good to go.

src/Guide/Api/Methods.hs
```diff
@@ -30,20 +31,20 @@ import qualified Guide.Search as Search
 ----------------------------------------------------------------------------
 
 -- | Get a list of available categories.
-getCategories :: DB -> Handler [CCategoryInfo]
+getCategories :: DB -> Guider [CCategoryInfo]
 getCategories db = do
   dbQuery db GetCategories <&> \xs ->
     map toCCategoryInfo xs
```
kana-sama, line 37: why do + <&>, maybe it is better try

```haskell
getCategories = do
  categories <- dbQuery db GetCategories
  pure $ map toCCCategoryInfo categories
```

willbasky: This code was before me. We can restyle all code in another PR. I opened some tasks at notion.

src/Guide/Api/Guider.hs
```diff
@@ -0,0 +1,39 @@
+{- | Guider monad with Config to replace servant`s Handler. -}
```
neongreen: How does it keep happening that you use `` ` `` instead of `'`?

```suggestion
{- | 'Guider' monad with 'Config' to replace Servant's 'Handler'. -}
```

neongreen: Also, I'm not sure whether this will work in Haddock, it's safer to move this comment to be directly before the module declaration.

src/Guide/Api/Guider.hs
```diff
@@ -0,0 +1,39 @@
...
+
+import Servant (Handler (..))
+import Servant.Server.Generic
+import Servant.Server.Internal.ServantErr (ServantErr)
+
+import Guide.Config (Config (..), def)
+
+
+-- | Guider type contains Config inside and replaces servant`s Handler type.
+newtype Guider a = Guider
```
neongreen, line 19: A comment should say _why_ something is being done, because _what_ is being done is usually the easy part (you can see it directly from the code). I.e.: why does `Guider` exist? What problem does it solve?

src/Guide/Api/Guider.hs
```diff
@@ -0,0 +1,39 @@
...
+  { runGuider :: ReaderT Config IO a
+  } deriving (Functor, Applicative, Monad, MonadIO, MonadReader Config)
+
+instance MonadError ServantErr Guider where
+  throwError :: ServantErr -> Guider a
+  throwError = liftIO . throwIO
+
+  catchError :: Guider a -> (ServantErr -> Guider a) -> Guider a
+  catchError (Guider m) f =
+    (Guider $ ReaderT $ \r -> (runReaderT m r)) `catchError`
```
neongreen: This definition calls itself and thus won't work.

src/Guide/Api/Guider.hs
```diff
@@ -0,0 +1,39 @@
...
+      (\e -> Guider $ ReaderT $ \r -> (runReaderT (runGuider (f e)) r))
+
+-- | Convertor from Guider to Handler
+guiderToHandler :: Guider a -> Handler a
+guiderToHandler (Guider m) = Handler $ ExceptT $ Right <$> runReaderT m def
```
neongreen: Nope, this is still wrong – it does the same thing as the definition before. The whole point is that you _don't_ want to always return `Right`. You want to return `Left` if a ServantErr has happened.

src/Guide/Api/Methods.hs
```diff
@@ -30,23 +31,23 @@ import qualified Guide.Search as Search
...
 -- | Create a new category, given the title and the grandparent (aka group).
 --
 -- Returns the ID of the created category (or of the existing one if the
 -- category with this title exists already).
-createCategory :: DB -> Text -> Text -> Handler (Uid Category)
+createCategory :: DB -> Text -> Text -> Guider (Uid Category)
 createCategory db title' group' = do
-  when (T.null title') $ do throwError (err400 {errBody = "Title not provided"})
-  when (T.null group') $ do throwError (err400 {errBody = "Group' not provided"})
+  when (T.null title') $ throwError err400 {errBody = "Title not provided"}
```
neongreen: If you decide to remove parentheses, perhaps it's better to make the coupling tighter so as not to confuse people.

```suggestion
  when (T.null title') $ throwError err400{errBody = "Title not provided"}
```

src/Guide/Api/Server.hs
```diff
@@ -71,13 +71,20 @@ type FullApi =
...
   where
     doc = toSwagger (Proxy @Api)
             & info.title   .~ "Aelve Guide API"
             & info.version .~ "alpha"
 
+-- | Convertor from api with Guider to api with Handler via hoistServer.
+api :: DB -> Server Api
+api db = hoistServer proxy guiderToHandler (toServant (guiderServer db))
+
+proxy :: Proxy Api
```
neongreen: You don't need this definition, just write `hoistServer (Proxy @Api) ...`. You'll need to enable `TypeApplications` for that.

src/Guide/Api/Guider.hs
```diff
@@ -0,0 +1,39 @@
...
+       ( Guider (..)
+       , GuiderServer
+       , guiderToHandler
+       ) where
+
+import Imports
+
+import Servant (Handler (..))
+import Servant.Server.Generic
+import Servant.Server.Internal.ServantErr (ServantErr)
```
neongreen: I think this is exported from `Servant`, no need to grasp for `Internal`.

src/Guide/Api/Guider.hs
```diff
@@ -0,0 +1,39 @@
...
+instance MonadError ServantErr Guider where
+  throwError :: ServantErr -> Guider a
+  throwError = liftIO . throwIO
+
+  catchError :: Guider a -> (ServantErr -> Guider a) -> Guider a
+  catchError (Guider m) f =
+    (Guider $ ReaderT $ \r -> (runReaderT m r)) `catchError`
+      (\e -> Guider $ ReaderT $ \r -> (runReaderT (runGuider (f e)) r))
+
+-- | Convertor from Guider to Handler
```
neongreen:
```suggestion
-- | Lift a 'Guider' handler into Servant's 'Handler'.
--
-- You can use this function with 'hoistServer' to run a 'GuiderServer'.
```

src/Guide/Api/Guider.hs
```diff
@@ -0,0 +1,36 @@
...
+  { runGuider :: ReaderT Config IO a
+  } deriving (Functor, Applicative, Monad, MonadIO, MonadReader Config)
+
+instance MonadError ServantErr Guider where
+  throwError :: ServantErr -> Guider a
+  throwError = liftIO . throwIO
+
+  catchError :: Guider a -> (ServantErr -> Guider a) -> Guider a
+  catchError (Guider m) f =
+    liftIO $ (runReaderT m def) `catch` (\e -> runReaderT (runGuider (f e)) def)
```
neongreen: "If it typechecks, it works" doesn't work here 🙂 

You can't ignore the reader argument and run the handler with `def` as the config. It would be buggy behavior.

Here's something that should work (I haven't tried it):

```haskell
  catchError (Guider m) f = Guider $ ReaderT $ \config ->
    runReaderT m config `catch` (\err -> runReaderT (runGuider (f err)) config)
```

src/Guide/Api/Guider.hs
```diff
@@ -0,0 +1,36 @@
...
+
+-- | The custom type won't be accepted by servant server without this conventor using at 'hoistServer'.
```
neongreen:
```suggestion
-- | The custom type won't be accepted by servant server without this convertor used with 'hoistServer'.
```

src/Guide/Api/Server.hs
```diff
@@ -52,39 +52,44 @@ apiServer db = Site
...
+api db config = hoistServer (Proxy @Api) (guiderToHandler config) (toServant $ guiderServer db)
+
 -- | Serve the API on port 4400.
 --
 -- You can test this API by doing @withDB mempty runApiServer@.
 runApiServer :: Config -> AcidState GlobalState -> IO ()
 runApiServer Config{..} db = do
   say $ format "API is running on port {}" _portApi
-  run _portApi $ corsPolicy $ serve (Proxy @FullApi) (fullServer db)
+  config <- readConfig
```
neongreen: The config is already passed to `runApiServer`.

neongreen (approved):

PR #246, opened 2018-11-26
Pin highlighting-kate

neongreen:
We pin the precise versions of 'highlighting-kate' and 'cmark-highlight'
because the frontend has copied a stylesheet from 'highlighting-kate'. If
the version changes, we also have to update the stylesheet on the
frontend. (See App.vue, or just grep for things like "span.st".)


neongreen:
Okay, this passes now, except for the frontend tests (but the failure there is unrelated).

willbasky (approved):

PR #248, opened 2018-11-29
Add addEdit

willbasky:
- [x] Instance RequestDetails
- [x] Integrate with api and server
- [x] AddEdit

### DIFF
--- a/src/Guide/Api/Methods.hs
+++ b/src/Guide/Api/Methods.hs
@@ -133,7 +133,7 @@ setItemInfo db requestDetails itemId CItemInfo{..} = do
   mapM_ (addEdit db requestDetails) [editName, editGroup, editLink, editKind]
   pure NoContent
 
--- | Set item's summary.SetItemKind itemId $ unH ciiKind
+-- | Set item's summary.
 setItemSummary :: DB -> RequestDetails -> Uid Item -> CTextEdit -> Guider NoContent
 setItemSummary db requestDetails itemId CTextEdit{..} = do
   serverModified <- markdownBlockMdSource . _itemDescription <$> getItemOrFail db itemId
@@ -163,6 +163,7 @@ setItemNotes db requestDetails itemId CTextEdit{..} = do
 -- | Delete an item.
 deleteItem :: DB -> RequestDetails -> Uid Item -> Guider NoContent
 deleteItem db requestDetails itemId = do
+  _ <- getItemOrFail db itemId
   dbUpdate db (DeleteItem itemId) >>= (mapM_ $ \edit -> do
     addEdit db requestDetails edit)
   pure NoContent
@@ -196,6 +197,7 @@ setTrait db requestDetails itemId traitId CTextEdit{..} = do
 -- | Delete a trait (pro/con).
 deleteTrait :: DB -> RequestDetails -> Uid Item -> Uid Trait -> Guider NoContent
 deleteTrait db requestDetails itemId traitId = do
+  _ <- getTraitOrFail db itemId traitId
   dbUpdate db (DeleteTrait itemId traitId) >>= (mapM_ $ \edit -> do
     addEdit db requestDetails edit)
   pure NoContent
--- a/src/Guide/Api/Methods.hs
+++ b/src/Guide/Api/Methods.hs
@@ -157,8 +157,8 @@ setItemNotes db itemId CTextEdit{..} = do
 deleteItem :: DB -> Uid Item -> Guider NoContent
 deleteItem db itemId = do
   _mbEdit <- dbUpdate db (DeleteItem itemId)
+--   mapM_ addEdit _mbEdit
   pure NoContent
-  -- TODO: mapM_ addEdit mbEdit
 
 ----------------------------------------------------------------------------
 -- Traits
@@ -190,7 +190,6 @@ deleteTrait :: DB -> Uid Item -> Uid Trait -> Guider NoContent
 deleteTrait db itemId traitId = do
   _mbEdit <- dbUpdate db (DeleteTrait itemId traitId)
   pure NoContent
-
 
 ----------------------------------------------------------------------------
 -- Search
--- a/src/Guide/Api/Methods.hs
+++ b/src/Guide/Api/Methods.hs
@@ -23,7 +23,6 @@ import Guide.Markdown (MarkdownBlock (..), MarkdownInline (..), MarkdownTree (..
 import Guide.State
 import Guide.Types
 import Guide.Utils
-import Data.IP (IP)
 
 import qualified Data.Set as S
 import qualified Data.Text as T
@@ -60,9 +59,8 @@ createCategory db requestDetails title' group' = do
     Nothing -> do
       catId <- randomShortUid
       time <- liftIO getCurrentTime
-      (_edit, _) <- dbUpdate db (AddCategory catId title' group' time)
-      config <- ask
-      addEdit db config requestDetails _edit
+      (edit, _) <- dbUpdate db (AddCategory catId title' group' time)
+      addEdit db requestDetails edit
       return catId
 
 -- | Edit categoty's note.
@@ -70,9 +68,8 @@ setCategoryNotes :: DB -> RequestDetails -> Uid Category -> CTextEdit -> Guider 
 setCategoryNotes db requestDetails catId CTextEdit{..} = do
   serverModified <- markdownBlockMdSource . _categoryNotes <$> getCategoryOrFail db catId
   checkConflict CTextEdit{..} serverModified
-  (_edit, _) <- dbUpdate db (SetCategoryNotes catId $ unH cteModified)
-  config <- ask
-  addEdit db config requestDetails _edit
+  (edit, _) <- dbUpdate db (SetCategoryNotes catId $ unH cteModified)
+  addEdit db requestDetails edit
   pure NoContent
 
 -- | Edit category's info (title, group, status, sections (pro/con, ecosystem, note)).
@@ -80,29 +77,24 @@ setCategoryInfo :: DB -> RequestDetails -> Uid Category -> CCategoryInfoEdit -> 
 setCategoryInfo db requestDetails catId CCategoryInfoEdit{..} = do
   category <- getCategoryOrFail db catId
   -- TODO diff and merge
-  (_editTitle, _) <- dbUpdate db $ SetCategoryTitle catId $ unH ccieTitle
-  (_editGroup, _) <- dbUpdate db $ SetCategoryGroup catId $ unH ccieGroup
-  (_editStatus, _) <- dbUpdate db $ SetCategoryStatus catId $ unH ccieStatus
+  (editTitle, _) <- dbUpdate db $ SetCategoryTitle catId $ unH ccieTitle
+  (editGroup, _) <- dbUpdate db $ SetCategoryGroup catId $ unH ccieGroup
+  (editStatus, _) <- dbUpdate db $ SetCategoryStatus catId $ unH ccieStatus
   let oldEnabledSections = category ^. enabledSections
   let newEnabledSections = unH ccieSections
-  (_editSection, _) <- dbUpdate db $ ChangeCategoryEnabledSections catId
+  (editSection, _) <- dbUpdate db $ ChangeCategoryEnabledSections catId
       (newEnabledSections S.\\ oldEnabledSections)
       (oldEnabledSections S.\\ newEnabledSections)
-  config <- ask
-  mapM_ (addEdit db config requestDetails) [_editTitle, _editGroup, _editStatus, _editSection]
+  mapM_ (addEdit db requestDetails) [editTitle, editGroup, editStatus, editSection]
   pure NoContent
 
 -- | Delete a category.
 deleteCategory :: DB -> RequestDetails -> Uid Category -> Guider NoContent
 deleteCategory db requestDetails catId = do
   _ <- getCategoryOrFail db catId
-  eitherEdit <- dbUpdate db (DeleteCategory catId)
-  case eitherEdit of
-    Left _ -> pure NoContent
-    Right _edit -> do
-      config <- ask
-      addEdit db config requestDetails _edit
-      pure NoContent
+  dbUpdate db (DeleteCategory catId) >>= (mapM_ $ \edit -> do
+    addEdit db requestDetails edit)
+  pure NoContent
 
 ----------------------------------------------------------------------------
 -- Items
@@ -123,9 +115,8 @@ createItem db requestDetails catId name' = do
       looksLikeLibrary = T.all isAllowedChar name'
       kind' = if looksLikeLibrary then Library (Just name') else Other
   time <- liftIO getCurrentTime
-  (_edit, _) <- dbUpdate db (AddItem catId itemId name' time kind')
-  config <- ask
-  addEdit db config requestDetails _edit
+  (edit, _) <- dbUpdate db (AddItem catId itemId name' time kind')
+  addEdit db requestDetails edit
   pure itemId
 
 -- TODO: move an item
@@ -135,22 +126,20 @@ setItemInfo :: DB -> RequestDetails -> Uid Item -> CItemInfo -> Guider NoContent
 setItemInfo db requestDetails itemId CItemInfo{..} = do
   _ <- getItemOrFail db itemId
   -- TODO diff and merge
-  (_editName, _) <- dbUpdate db $ SetItemName itemId $ unH ciiName
-  (_editGroup, _) <- dbUpdate db $ SetItemGroup itemId $ unH ciiGroup
-  (_editLink, _) <- dbUpdate db $ SetItemLink itemId $ unH ciiLink
-  (_editKind, _) <- dbUpdate db $ SetItemKind itemId $ unH ciiKind
-  config <- ask
-  mapM_ (addEdit db config requestDetails) [_editName, _editGroup, _editLink, _editKind]
+  (editName, _) <- dbUpdate db $ SetItemName itemId $ unH ciiName
+  (editGroup, _) <- dbUpdate db $ SetItemGroup itemId $ unH ciiGroup
+  (editLink, _) <- dbUpdate db $ SetItemLink itemId $ unH ciiLink
+  (editKind, _) <- dbUpdate db $ SetItemKind itemId $ unH ciiKind
+  mapM_ (addEdit db requestDetails) [editName, editGroup, editLink, editKind]
   pure NoContent
 
--- | Set item's summary.
+-- | Set item's summary.SetItemKind itemId $ unH ciiKind
 setItemSummary :: DB -> RequestDetails -> Uid Item -> CTextEdit -> Guider NoContent
 setItemSummary db requestDetails itemId CTextEdit{..} = do
   serverModified <- markdownBlockMdSource . _itemDescription <$> getItemOrFail db itemId
   checkConflict CTextEdit{..} serverModified
-  (_edit, _) <- dbUpdate db (SetItemDescription itemId $ unH cteModified)
-  config <- ask
-  addEdit db config requestDetails _edit
+  (edit, _) <- dbUpdate db (SetItemDescription itemId $ unH cteModified)
+  addEdit db requestDetails edit
   pure NoContent
 
 -- | Set item's ecosystem.
@@ -158,9 +147,8 @@ setItemEcosystem :: DB -> RequestDetails -> Uid Item -> CTextEdit -> Guider NoCo
 setItemEcosystem db requestDetails itemId CTextEdit{..} = do
   serverModified <- markdownBlockMdSource . _itemEcosystem <$> getItemOrFail db itemId
   checkConflict CTextEdit{..} serverModified
-  (_edit, _) <- dbUpdate db (SetItemEcosystem itemId $ unH cteModified)
-  config <- ask
-  addEdit db config requestDetails _edit
+  (edit, _) <- dbUpdate db (SetItemEcosystem itemId $ unH cteModified)
+  addEdit db requestDetails edit
   pure NoContent
 
 -- | Set item's notes.
@@ -168,21 +156,16 @@ setItemNotes :: DB -> RequestDetails -> Uid Item -> CTextEdit -> Guider NoConten
 setItemNotes db requestDetails itemId CTextEdit{..} = do
   serverModified <- markdownTreeMdSource . _itemNotes <$> getItemOrFail db itemId
   checkConflict CTextEdit{..} serverModified
-  (_edit, _) <- dbUpdate db (SetItemNotes itemId $ unH cteModified)
-  config <- ask
-  addEdit db config requestDetails _edit
+  (edit, _) <- dbUpdate db (SetItemNotes itemId $ unH cteModified)
+  addEdit db requestDetails edit
   pure NoContent
 
 -- | Delete an item.
 deleteItem :: DB -> RequestDetails -> Uid Item -> Guider NoContent
 deleteItem db requestDetails itemId = do
-  eitherEdit <- dbUpdate db (DeleteItem itemId)
-  case eitherEdit of
-    Left _ -> pure NoContent
-    Right _edit -> do
-      config <- ask
-      addEdit db config requestDetails _edit
-      pure NoContent
+  dbUpdate db (DeleteItem itemId) >>= (mapM_ $ \edit -> do
+    addEdit db requestDetails edit)
+  pure NoContent
 
 ----------------------------------------------------------------------------
 -- Traits
@@ -195,11 +178,10 @@ createTrait :: DB -> RequestDetails -> Uid Item -> TraitType -> Text -> Guider (
 createTrait db requestDetails itemId traitType text = do
   when (T.null text) $ throwError err400{errBody = "Trait text not provided"}
   traitId <- randomShortUid
-  (_edit, _) <- case traitType of
+  (edit, _) <- case traitType of
     Con -> dbUpdate db (AddCon itemId traitId text)
     Pro -> dbUpdate db (AddPro itemId traitId text)
-  config <- ask
-  addEdit db config requestDetails _edit
+  addEdit db requestDetails edit
   pure traitId
 
 -- | Update the text of a trait (pro/con).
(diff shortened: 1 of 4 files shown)


src/Guide/Api/Methods.hs
```diff
@@ -157,8 +157,8 @@ setItemNotes db itemId CTextEdit{..} = do
 deleteItem :: DB -> Uid Item -> Guider NoContent
 deleteItem db itemId = do
   _mbEdit <- dbUpdate db (DeleteItem itemId)
+--   mapM_ addEdit _mbEdit
```
kana-sama: What is it? Why at first it is commented and then in todo?

src/Guide/Api/Utils.hs
```diff
@@ -89,3 +94,43 @@ instance (HasSwagger api, KnownSymbol name, KnownSymbol desc) =>
...
             Nothing
      in toSwagger (Proxy @api) & applyTags [tag]
+
+
+-- | Type to log user's edition with header details
+data RequestDetails = RequestDetails
+  { rdIp      :: Text
+  , rdReferer :: Text
+  , rdUA      :: Text
+  } deriving (Show, Generic)
```
kana-sama, line 112: please add docs for fields, I don't know, what rdUA is about.

Or maybe just more meaningful names like rdUserAgent

src/Guide/Api/Utils.hs
```diff
@@ -89,3 +94,43 @@ instance (HasSwagger api, KnownSymbol name, KnownSymbol desc) =>
...
+  hoistServerWithContext _ pc nt s = hoistServerWithContext (Proxy :: Proxy api) pc nt . s
+
+  route Proxy context subserver = route (Proxy :: Proxy api) context $
+      subserver `addHeaderCheck` withRequest headerCheck
+    where
+      headerCheck :: Request -> DelayedIO (Maybe RequestDetails)
+      headerCheck req = unfoldRequestArgument (Proxy :: Proxy '[Optional, Strict]) errReq errSt meRequestDatails
+        where
+          meRequestDatails :: Maybe (Either Text RequestDetails)
+          meRequestDatails = getCompose $ liftA2 RequestDetails (Compose ip) (Compose referer) <*> Compose ua
```
kana-sama: `liftA2 f a b <*> c` = `liftA3 f a b c`

src/Guide/Api/Utils.hs
```diff
@@ -89,3 +94,43 @@ instance (HasSwagger api, KnownSymbol name, KnownSymbol desc) =>
...
+
+          ip :: FromHttpApiData a => Maybe (Either Text a)
+          ip = fmap parseHeader $ lookup "X-Forwarded-For" (requestHeaders req)
+          referer :: FromHttpApiData a => Maybe (Either Text a)
+          referer = fmap parseHeader $ lookup hReferer (requestHeaders req)
+          ua :: FromHttpApiData a => Maybe (Either Text a)
+          ua = fmap parseHeader $ lookup hUserAgent (requestHeaders req)
```
kana-sama: Maybe we case make something like:
```haskell
parseHeader :: String -> Maybe (Either Text a)
parseHeader key = lookup key (requestHeader req)

ip = parseHeader "X-Forwarded-For"
referer = parseHeader hReferer
ua = parseHeader hUserAgent
```

Also you can generalize RequestDetails over Text and then derive Traversable instance, so you just can write:
```haskell
meRequestDatails = getCompose $ traverse (Compose . parseHeader)
  (RequestDetails "X-Forwarded-For" hReferer hUserAgent)
```

kana-sama (changes requested):

src/Guide/Api/Utils.hs
kana-sama: Simple example:

```haskell
{-# LANGUAGE DeriveFunctor, DeriveFoldable, DeriveTraversable #-}

import Data.Functor.Compose

data Error = Error
  deriving Show

data RequestDetails' a = RequestDetails
  { ip :: a
  , referer :: a
  , userAgent :: a
  } deriving (Functor, Foldable, Traversable, Show)

type RequestDetails = RequestDetails' String

parseHeader :: String -> Maybe (Either Error String)
-- parseHeader "ip" = Just (Left Error)
-- parseHeader "ip" = Nothing
parseHeader key = Just (Right key)

requestDetails :: Maybe (Either Error RequestDetails)
requestDetails = getCompose $ traverse (Compose . parseHeader) $ RequestDetails
  { ip = "ip"
  , referer = "referer"
  , userAgent = "userAgent"
  }
```

src/Guide/Api/Methods.hs
neongreen: It was commented so that we wouldn't forget to add it later.

I'm not sure why this line was changed in this PR, though.

src/Guide/Api/Utils.hs
```diff
@@ -89,3 +94,43 @@ instance (HasSwagger api, KnownSymbol name, KnownSymbol desc) =>
...
+
+          errReq = delayedFailFatal err400
+            { errBody = "One of Headers: " <> "Referer, X-Forwarded-For or User-Agent" <> " is required" }
```
kana-sama: Is it a case for 404? There are no like "Not found", it is just a validation error and it should be 400

src/Guide/Api/Utils.hs
```diff
@@ -89,3 +94,43 @@ instance (HasSwagger api, KnownSymbol name, KnownSymbol desc) =>
              symbolVal (Proxy @desc))
             Nothing
      in toSwagger (Proxy @api) & applyTags [tag]
+
+
+-- | Type to log user's edition with header details
```
neongreen: Well it's not necessarily to log edits, is it? It's just a general-purpose thing to get request details. We might well end up using it for more stuff.

```suggestion
-- | Servant request details, can be captured by adding `RequestDetails :>` to an API branch.
```

src/Guide/Api/Utils.hs
```diff
@@ -89,3 +94,43 @@ instance (HasSwagger api, KnownSymbol name, KnownSymbol desc) =>
...
+data RequestDetails = RequestDetails
+  { rdIp      :: Text
+  , rdReferer :: Text
+  , rdUA      :: Text
+  } deriving (Show, Generic)
+
+instance (HasServer api context)
+  => HasServer (Maybe RequestDetails :> api) context where
+
+  type ServerT (Maybe RequestDetails :> api) m = Maybe RequestDetails -> ServerT api m
```
neongreen: I get the point of `Maybe` here but it's rather unusual for a Servant combinator.

It's better to have just `RequestDetails` and make the fields optional. For some handlers we might need only some fields, for instance, and for some handlers we might need all of them. The handler should decide.

src/Guide/Api/Utils.hs
neongreen: I would advise against doing this because it relies on all fields having the same type. If we decide to parse IPs, for instance, it will get uglier.

src/Guide/Api/Utils.hs
neongreen: It shouldn't even be 400, it should just result in a `Nothing`. We definitely _definitely_ don't want to throw errors of any kind for requests that don't have a `Referer`. This will break curl, this will break people who intentionally hide the referer, etc.

src/Guide/Api/Utils.hs
```diff
@@ -89,3 +94,43 @@ instance (HasSwagger api, KnownSymbol name, KnownSymbol desc) =>
...
+          referer = fmap parseHeader $ lookup hReferer (requestHeaders req)
+          ua :: FromHttpApiData a => Maybe (Either Text a)
+          ua = fmap parseHeader $ lookup hUserAgent (requestHeaders req)
+
+          errReq = delayedFailFatal err400
+            { errBody = "One of Headers: " <> "Referer, X-Forwarded-For or User-Agent" <> " is required" }
+          errSt e = delayedFailFatal err400
+            { errBody = "Error parsing one of headers "
+           <> "Referer, X-Forwarded-For or User-Agent"
+           <> " failed: "
```
kana-sama: List of headers can be generalized over error messages. 

ALSO, it is a place for future mistake, when we will add a field to RD, but forget to add it here, so there is another great typecheckable solution, whih is based on previous solution: 

```haskell
{-# LANGUAGE DeriveFunctor, DeriveFoldable, DeriveTraversable #-}

import Data.Functor.Compose
import Data.Foldable
import Data.List (intercalate)

data Error = Error
  deriving Show

type RequestDetails = RequestDetails' String
data RequestDetails' a = RequestDetails
  { ip :: a
  , referer :: a
  , userAgent :: a
  } deriving (Functor, Foldable, Traversable, Show)

-- Just for visibility
type HeaderKey = String

parseHeader :: HeaderKey -> Maybe (Either Error String)
-- parseHeader "ip" = Just (Left Error)
-- parseHeader "ip" = Nothing
parseHeader key = Just (Right key)

headersMap :: RequestDetails' HeaderKey
headersMap = RequestDetails
  { ip = "X-Forwarded-For"
  , referer = "Referer"
  , userAgent = "User-Agent"
  }

requestDetails :: Maybe (Either Error RequestDetails)
requestDetails = getCompose $ traverse (Compose . parseHeader) headersMap
  
headers :: String
headers = intercalate ", " (toList headersMap)

err :: String
err = "One of Headers: " <> headers <> " is required"
```

kana-sama (changes requested):

src/Guide/Api/Utils.hs
neongreen: We can't forget to add the field – `liftA3` will stop typechecking.

kana-sama: There are no liftA3 in headers list, it is just a list of string

neongreen: Ah, you mean the `"Referer, X-Forwarded-For or User-Agent"` thingy?

It should just go away completely once `Maybe` is pushed down into the field types.

src/Guide/Api/Utils.hs
kana-sama: We can rewrite it in future to liftA3 when it will be needed, there are no reasons not to do this right now. Rewriting shouldn't be difficult. Using polymorphic RequestDetails' in other modules can be the only problem (we can solve this with hiding polymorphic type and exporting just lenses (not polymorphic)).

It is a simple GHC code generation for a simple task. When task will become complex, the solution will become complex too.

neongreen: The reason it should be done right now is that the next thing I'm going to ask for in this PR is "please make the IP field have the type `IP`" 🙂

neongreen: By the way, the IP determining logic is more complicated than just taking the Forwarded-For header.

```haskell
  mbForwardedFor <- liftA2 (<|>) (Spock.header "Forwarded-For")
                                 (Spock.header "X-Forwarded-For")
  mbIP <- case mbForwardedFor of
    Nothing -> sockAddrToIP . Wai.remoteHost <$> Spock.request
    Just ff -> case readMaybe (toString ip) of
      Nothing -> error ("couldn't read Forwarded-For address: " ++
                        show ip ++ " (full header: " ++
                        show ff ++ ")")
      Just i  -> return (Just i)
      where
        addr = T.strip . snd . T.breakOnEnd "," $ ff
        ip -- [IPv6]:port
           | T.take 1 addr == "[" =
               T.drop 1 (T.takeWhile (/= ']') addr)
           -- IPv4 or IPv4:port
           | T.any (== '.') addr =
               T.takeWhile (/= ':') addr
           -- IPv6 without port
           | otherwise =
               addr
```

The "error" here is not nice, but everything else should be ported.

kana-sama: Ok, it makes sense. it's a pity, I love Traversable

kana-sama: Ok, it makes sense. It's a pity, I love Traversable.

willbasky:
Why it is not _IP_ in `ipFromHost :: Maybe Text`?
Because it will return a demand on `instance FromHttpApiData IP` for _hoistServer_ in _Server_ module later.

src/Guide/Api/Methods.hs
```diff
@@ -57,41 +60,49 @@ createCategory db title' group' = do
...
+deleteCategory :: DB -> RequestDetails -> Uid Category -> Guider NoContent
+deleteCategory db requestDetails catId = do
   _ <- getCategoryOrFail db catId
-  _mbEdit <- dbUpdate db (DeleteCategory catId)
-  pure NoContent
-  -- TODO mapM_ addEdit mbEdit
+  eitherEdit <- dbUpdate db (DeleteCategory catId)
+  case eitherEdit of
+    Left _ -> pure NoContent
+    Right _edit -> do
```
neongreen: The reason for the `_` prefix in variable names is that it doesn't trip the unused variable checker.

If you use the variable, `_` is not needed.

src/Guide/Api/Methods.hs
```diff
@@ -112,53 +123,66 @@ createItem db catId name' = do
...
-deleteItem :: DB -> Uid Item -> Guider NoContent
-deleteItem db itemId = do
-  _mbEdit <- dbUpdate db (DeleteItem itemId)
-  pure NoContent
-  -- TODO: mapM_ addEdit mbEdit
+deleteItem :: DB -> RequestDetails -> Uid Item -> Guider NoContent
+deleteItem db requestDetails itemId = do
+  eitherEdit <- dbUpdate db (DeleteItem itemId)
+  case eitherEdit of
+    Left _ -> pure NoContent
```
neongreen: It should also fail, not just return `NoContent`.

src/Guide/Api/Methods.hs
```diff
@@ -221,6 +252,16 @@ dbQuery :: (MonadIO m, EventState event ~ GlobalState, QueryEvent event)
...
   Acid.query db x
 
+-- Call this whenever any user-made change is applied to the database.
+addEdit :: MonadIO m => DB -> Config -> RequestDetails -> Edit -> m ()
+addEdit db Config{..} RequestDetails{..} edit = unless (isVacuousEdit edit) $ do
+    time <- liftIO getCurrentTime
+    dbUpdate db $ RegisterEdit edit ip time
+    dbUpdate db $ RegisterAction (Action'Edit edit) ip time _baseUrl rdReferer rdUserAgent
+  where
+    ip :: Maybe IP
```
neongreen: Hm, how will it work? `read @(Maybe IP)` is going to parse strings like "Just 251.35.23.186". Maybe you wanted to use `readMaybe`?

src/Guide/Api/Utils.hs
```diff
@@ -89,3 +99,42 @@ instance (HasSwagger api, KnownSymbol name, KnownSymbol desc) =>
...
+
+  hoistServerWithContext _ pc nt s = hoistServerWithContext (Proxy :: Proxy api) pc nt . s
+
+  route Proxy context subserver = route (Proxy :: Proxy api) context $
+      subserver `addHeaderCheck` withRequest getRequestDetails
+    where
+      getRequestDetails :: Request -> DelayedIO RequestDetails
+      getRequestDetails req = pure $ RequestDetails ip referer useragent
+        where
+          ip :: FromHttpApiData Text => Maybe Text
```
neongreen: Why do you need this constraint? Shouldn't it be in scope already?

neongreen:
> Because it will return a demand on instance FromHttpApiData IP for hoistServer in Server module later.

Well, then just write an instance of FromHttpApiData IP 🙂

src/Guide/Api/Methods.hs
```diff
@@ -57,41 +60,49 @@ createCategory db title' group' = do
...
 -- | Delete a category.
-deleteCategory :: DB -> Uid Category -> Guider NoContent
-deleteCategory db catId = do
+deleteCategory :: DB -> RequestDetails -> Uid Category -> Guider NoContent
+deleteCategory db requestDetails catId = do
   _ <- getCategoryOrFail db catId
-  _mbEdit <- dbUpdate db (DeleteCategory catId)
-  pure NoContent
-  -- TODO mapM_ addEdit mbEdit
+  eitherEdit <- dbUpdate db (DeleteCategory catId)
```
kana-sama: I prefer LambdaCase (because of temporal variables) and

```haskell
dbUpdate db (DeleteCategory catId) >>= \case
  Left _ -> pure NoContent
  Right edit' -> do
    config <- ask
    addEdit db config requestDetails edit'
    pure NoContent
```

but for this case there is another way:
```haskell
dbUpdate db (DeleteCategory catId) >>= mapM $ \edit' ->
  config <- ask
  addEdit db config requestDetails edit'
pure NoContent
```

src/Guide/Api/Methods.hs
```diff
@@ -112,53 +123,66 @@ createItem db catId name' = do
...
   _ <- getItemOrFail db itemId
   -- TODO diff and merge
-  _ <- dbUpdate db $ SetItemName itemId $ unH ciiName
-  _ <- dbUpdate db $ SetItemGroup itemId $ unH ciiGroup
-  _ <- dbUpdate db $ SetItemLink itemId $ unH ciiLink
-  _ <- dbUpdate db $ SetItemKind itemId $ unH ciiKind
+  (_editName, _) <- dbUpdate db $ SetItemName itemId $ unH ciiName
+  (_editGroup, _) <- dbUpdate db $ SetItemGroup itemId $ unH ciiGroup
+  (_editLink, _) <- dbUpdate db $ SetItemLink itemId $ unH ciiLink
+  (_editKind, _) <- dbUpdate db $ SetItemKind itemId $ unH ciiKind
```
kana-sama: Can be generalized, but I don't know, should we
```haskell
let edits =
  [ SetItemLink itemId $ unH ciiLink
  , SetItemGroup itemId $ unH ciiGroup
  , SetItemLink itemId $ unH ciiLink
  , SetItemKind itemId $ unH ciiKind
  ]
edits' <- mapM (fmap fst . dbUpdate db) edits
mapM_ (addEdit db config requestDetails) edits'
```

Because I don't know, why there is a list?

src/Guide/Api/Methods.hs
```diff
@@ -112,53 +123,66 @@ createItem db catId name' = do
...
-setItemSummary :: DB -> Uid Item -> CTextEdit -> Guider NoContent
-setItemSummary db itemId CTextEdit{..} = do
+setItemSummary :: DB -> RequestDetails -> Uid Item -> CTextEdit -> Guider NoContent
+setItemSummary db requestDetails itemId CTextEdit{..} = do
   serverModified <- markdownBlockMdSource . _itemDescription <$> getItemOrFail db itemId
   checkConflict CTextEdit{..} serverModified
-  (_edit, _newItem) <- dbUpdate db (SetItemDescription itemId $ unH cteModified)
+  (_edit, _) <- dbUpdate db (SetItemDescription itemId $ unH cteModified)
+  config <- ask
+  addEdit db config requestDetails _edit
```
kana-sama: Maybe we can add Config context to the addEdit scope too? There are too many of asks.

src/Guide/Api/Methods.hs
willbasky: `edits` list is heterogeneous, therefore it needed to add `dbUpdate db $ `. And it become
```haskell
let edits =
        [ fmap fst $ dbUpdate db $ SetItemName itemId $ unH ciiName
        , fmap fst $ dbUpdate db $ SetItemGroup itemId $ unH ciiGroup
        , fmap fst $ dbUpdate db $ SetItemLink itemId $ unH ciiLink
        , fmap fst $ dbUpdate db $ SetItemKind itemId $ unH ciiKind
        ]
```
Not so elegant.

willbasky: > Because I don't know, why there is a list?

The list is to apply `addEdit` function.

src/Guide/Api/Methods.hs
```diff
@@ -112,53 +115,57 @@ createItem db catId name' = do
...
+  (editGroup, _) <- dbUpdate db $ SetItemGroup itemId $ unH ciiGroup
+  (editLink, _) <- dbUpdate db $ SetItemLink itemId $ unH ciiLink
+  (editKind, _) <- dbUpdate db $ SetItemKind itemId $ unH ciiKind
+  mapM_ (addEdit db requestDetails) [editName, editGroup, editLink, editKind]
   pure NoContent
 
--- | Set item's summary.
-setItemSummary :: DB -> Uid Item -> CTextEdit -> Guider NoContent
-setItemSummary db itemId CTextEdit{..} = do
+-- | Set item's summary.SetItemKind itemId $ unH ciiKind
```
neongreen: hm?

src/Guide/Api/Methods.hs
```diff
@@ -112,53 +115,57 @@ createItem db catId name' = do
...
+  addEdit db requestDetails edit
   pure NoContent
 
 -- | Delete an item.
-deleteItem :: DB -> Uid Item -> Guider NoContent
-deleteItem db itemId = do
-  _mbEdit <- dbUpdate db (DeleteItem itemId)
+deleteItem :: DB -> RequestDetails -> Uid Item -> Guider NoContent
+deleteItem db requestDetails itemId = do
+  dbUpdate db (DeleteItem itemId) >>= (mapM_ $ \edit -> do
```
neongreen, line 167: This should also check for item existence, like `deleteCategory` does.

src/Guide/Api/Methods.hs
```diff
@@ -167,31 +174,32 @@ deleteItem db itemId = do
...
+  addEdit db requestDetails edit
   pure NoContent
 
 -- | Delete a trait (pro/con).
-deleteTrait :: DB -> Uid Item -> Uid Trait -> Guider NoContent
-deleteTrait db itemId traitId = do
-  _mbEdit <- dbUpdate db (DeleteTrait itemId traitId)
+deleteTrait :: DB -> RequestDetails -> Uid Item -> Uid Trait -> Guider NoContent
+deleteTrait db requestDetails itemId traitId = do
+  dbUpdate db (DeleteTrait itemId traitId) >>= (mapM_ $ \edit -> do
```
neongreen, line 201: And this should check for trait existence.

src/Guide/Api/Utils.hs
```diff
@@ -89,3 +102,47 @@ instance (HasSwagger api, KnownSymbol name, KnownSymbol desc) =>
              symbolVal (Proxy @desc))
             Nothing
      in toSwagger (Proxy @api) & applyTags [tag]
+
+-- | Servant request details, can be captured by adding `RequestDetails :>` to an API branch.
+data RequestDetails = RequestDetails
+  { rdIp        :: Maybe IP  -- request ip address
```
neongreen: Field comments have to start with `-- ^`, otherwise they won't be rendered by Haddock.

src/Guide/Api/Utils.hs
```diff
@@ -89,3 +102,47 @@ instance (HasSwagger api, KnownSymbol name, KnownSymbol desc) =>
              symbolVal (Proxy @desc))
             Nothing
      in toSwagger (Proxy @api) & applyTags [tag]
+
+-- | Servant request details, can be captured by adding `RequestDetails :>` to an API branch.
```
neongreen: The syntax for code in Haddock is `@...@`, not `` `...` ``.

src/Guide/Api/Utils.hs
```diff
@@ -89,3 +102,47 @@ instance (HasSwagger api, KnownSymbol name, KnownSymbol desc) =>
...
+data RequestDetails = RequestDetails
+  { rdIp        :: Maybe IP  -- request ip address
+  , rdReferer   :: Maybe Text  -- request referrer
+  , rdUserAgent :: Maybe Text  -- request User-Agent
+  } deriving (Show, Generic)
+
+instance FromHttpApiData IP where
+  parseUrlPiece text
```
neongreen: Just use `readEither` instead, it might give a better error message instead of a generic "not a parseable..."

src/Guide/Api/Utils.hs
```diff
@@ -89,3 +102,47 @@ instance (HasSwagger api, KnownSymbol name, KnownSymbol desc) =>
...
+  hoistServerWithContext _ pc nt s = hoistServerWithContext (Proxy :: Proxy api) pc nt . s
+
+  route Proxy context subserver = route (Proxy :: Proxy api) context $
+      subserver `addHeaderCheck` withRequest getRequestDetails
+    where
+      getRequestDetails :: Request -> DelayedIO RequestDetails
+      getRequestDetails req = pure $ RequestDetails ip referer useragent
+        where
+          ip :: Maybe IP
+          ip = maybe ipFromHost Just $ (dropEither "Forwarded-For") <|> (dropEither "X-Forwarded-For")
```
neongreen: Try this (not sure if it will work):

```suggestion
          ip = ipFromHost <|> dropEither "Forwarded-For" <|> dropEither "X-Forwarded-For"
```

src/Guide/Api/Utils.hs
```diff
@@ -89,3 +102,47 @@ instance (HasSwagger api, KnownSymbol name, KnownSymbol desc) =>
...
+          referer :: Maybe Text
+          referer = dropEither hReferer
+          useragent :: Maybe Text
+          useragent = dropEither hUserAgent
+
+          dropEither :: FromHttpApiData a => NHTH.HeaderName -> Maybe a
```
neongreen: Why is it called `dropEither`? It gets a header; `getHeader` is a better name. (Or pick something else.)

src/Guide/Api/Utils.hs
neongreen: Also, it's not enough to just `read` the IP here. I showed this piece of code before:

```haskell
        addr = T.strip . snd . T.breakOnEnd "," $ forwardedForHeader

        ip = if
           -- [IPv6]:port
           | T.take 1 addr == "[" -> T.drop 1 (T.takeWhile (/= ']') addr)
           -- IPv4 or IPv4:port
           | T.any (== '.') addr  -> T.takeWhile (/= ':') addr
           -- IPv6 without port
           | otherwise            -> addr
```

I.e. you should create an additional helper for parsing the IP out of the `Forwarded-For` header.

src/Guide/Api/Utils.hs
```diff
@@ -89,3 +103,61 @@ instance (HasSwagger api, KnownSymbol name, KnownSymbol desc) =>
...
+
+          getHeader :: FromHttpApiData a => NHTH.HeaderName -> Maybe a
+          getHeader headerName = join $
+            (either (\_ -> Nothing) Just) . parseHeader <$> lookupName headerName
+
+          getIp :: Maybe ByteString -> Maybe IP
+          getIp mBody = case mBody of
+            Nothing -> sockAddrToIP $ remoteHost req
+            Just ff -> case readMaybe (toString ip) of
+              Nothing -> error $ "couldn't read Forwarded-For address: " ++
```
neongreen: Oh, this shouldn't error out – it should return `Nothing`. Yeah, my original code errored out, but that was a mistake.

src/Guide/Api/Utils.hs
```diff
@@ -89,3 +103,60 @@ instance (HasSwagger api, KnownSymbol name, KnownSymbol desc) =>
...
+          lookupName headerName = lookup headerName (requestHeaders req)
+
+          getHeader :: FromHttpApiData a => NHTH.HeaderName -> Maybe a
+          getHeader headerName = join $
+            (either (\_ -> Nothing) Just) . parseHeader <$> lookupName headerName
+
+          getIp :: Maybe ByteString -> Maybe IP
+          getIp mBody = case mBody of
+            Nothing -> sockAddrToIP $ remoteHost req
+            Just ff -> case readMaybe (toString ip) of
```
neongreen: Looks like now it's just `Just ff -> readMaybe (toString ip)`.

neongreen (approved):

PR #252, opened 2018-12-26
Push a Docker image after a Travis build

neongreen:
Left to fix: `js/bundle.js` is not copied into the image for some reason.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,8 +36,11 @@ back/run:
 # Docker file)
 .PHONY: back/travis-docker
 back/travis-docker:
-	git clone --depth 1 https://github.com/aelve/guide-database.git docker/back/state
-	rm -rf docker/back/state/.git
-	cp .stack-work/install/*/*/*/bin/guide docker/back/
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -30,3 +32,17 @@ back/test-db: back
 .PHONY: back/run
 back/run:
 	stack exec --cwd back -- guide
+
+# Create a Docker image for the backend; will only work on Travis because
+# the binary has to have been compiled on Ubuntu Xenial (the OS used in the
+# Docker file)
+.PHONY: back/travis-docker
```
willbasky, line 39: If I run this command localy?

neongreen: It won't work unless you have Ubuntu Xenial. The reason for that is that it copies a locally built binary into the Docker image (which uses Ubuntu) and this binary has to be compatible with system libraries.

neongreen: A better way to do this would be building everything inside the container on Travis, but this can be fixed later.

willbasky (approved):

PR #254, opened 2019-01-05
Move item or trait

willbasky:


back/src/Guide/Api/Methods.hs
```diff
@@ -163,6 +163,14 @@ deleteItem db requestDetails itemId = do
     addEdit db requestDetails edit)
   pure NoContent
 
+-- | Move item upper or lower
```
neongreen: "upper" and "lower" are not quite synonymical to "up" and "down".

You can use "up" and "down" everywhere, and for constructor names you can have `data Direction = DirectionUp | DirectionDown`.

back/src/Guide/Api/Types.hs
```diff
@@ -196,6 +197,15 @@ data ItemSite route = ItemSite
...
       :> Capture "id" (Uid Item)
       :> Delete '[JSON] NoContent
+
+  , _moveItem :: route :-
+      Summary "Move item"
+      :> ErrorResponse 404 "Item not found"
+      :> "item"
+      :> Capture "item" (Uid Item)
+      :> "move"
+      :> Capture "type" Move
```
neongreen: This should be in request body. You can create a separate type for it:

```haskell
data MoveRequest = MoveRequest
  { moveRequestDirection :: Direction
  } deriving (Show, Eq)

instance FromJSON MoveRequest where ...
```

back/src/Guide/Api/Types.hs
```diff
@@ -196,6 +197,15 @@ data ItemSite route = ItemSite
...
+      :> Put '[JSON] NoContent
```
neongreen: This should be a `Post`, not `Put`. The rule of thumb is that a `Put` request should always be safe to repeat, i.e. doing a `Put` request twice should be the same as doing it once.

back/src/Guide/Api/Types.hs
```diff
@@ -275,6 +297,27 @@ instance FromHttpApiData TraitType where
         "con" -> Right Con
         _     -> Left "Invalid trait type!"
 
+-- | Move (Upper/Lower) item or trait and their instances.
+data Move = Upper | Lower -- 'Down' is busy.
+    deriving (Eq, Show, Generic)
+
+instance ToSchema Move where
+    declareNamedSchema = genericDeclareNamedSchema schemaOptions
```
neongreen: The schema should specify that the acceptable values are "up" and "down".

back/src/Guide/Api/Types.hs
```diff
@@ -275,6 +297,27 @@ instance FromHttpApiData TraitType where
...
+
+instance ToParamSchema Move where
+    toParamSchema _ = mempty
+        & S.type_  .~ SwaggerString
+        & S.format ?~ "Move"
+
+instance ToHttpApiData Move where
+    toUrlPiece = toText . map toLower . show
```
neongreen: This generates "upper" and "lower", not "up" and "down".

back/src/Guide/Api/Methods.hs
```diff
@@ -163,20 +163,28 @@ deleteItem db requestDetails itemId = do
...
```
neongreen:
```suggestion
-- | Move item up or down
```

back/src/Guide/Api/Methods.hs
```diff
@@ -197,6 +205,14 @@ deleteTrait db requestDetails itemId traitId = do
     addEdit db requestDetails edit)
   pure NoContent
 
+-- | Move trait upper or lower
```
neongreen:
```suggestion
-- | Move trait up or down
```

back/src/Guide/Api/Types.hs
```diff
@@ -33,7 +33,10 @@ module Guide.Api.Types
   , CSearchResult(..), toCSearchResult
 
   -- * Other types
+  , Move(..)
+  , Direction(..)
   , TraitType(..)
+  , CreateNewTrait(..)
```
neongreen: Since all of these are client types, let's keep it consistent and add prefixes: `CMove`, `CDirection`, `CTraitType`, etc. Then you can also have `CCreateTrait` instead of `CreateNewTrait`, which would be more consistent since the method is called `createTrait`.

back/src/Guide/Api/Types.hs
```diff
@@ -261,19 +284,32 @@ data TraitType = Pro | Con
...
+    { constructorTagModifier = \case
+        "DirectionUp" -> "up"
+        "DirectionDown" -> "down"
+        other -> error ("Direction schema: unknown tag " <> show other)
+    }
 
-instance ToHttpApiData TraitType where
-    toUrlPiece = toText . map toLower . show
+instance A.ToJSON Direction where
+  toJSON = A.toJSON . drop (length ("Direction" :: String)) . show
```
neongreen: Won't work, you'll get `"Up"` and `"Down"` instead of `"up"` and `"down"`.

It's easier to just have a case statement with two branches:

```haskell
instance A.ToJSON Direction where
  toJSON = \case
    DirectionUp -> "up"
    DirectionDown -> "down"
```

back/src/Guide/Api/Types.hs
```diff
@@ -261,19 +284,32 @@ data TraitType = Pro | Con
...
-instance FromHttpApiData TraitType where
-    parseUrlPiece t = case t of
-        "pro" -> Right Pro
-        "con" -> Right Con
-        _     -> Left "Invalid trait type!"
+instance A.FromJSON Direction where
+  parseJSON = \case
+    "up"   -> pure DirectionUp
+    "down" -> pure DirectionDown
+    tag    -> fail ("unknown tag " ++ show tag)
```
neongreen:
```suggestion
    tag    -> fail ("unknown direction " ++ show tag)
```

back/src/Guide/Api/Types.hs
```diff
@@ -93,7 +96,7 @@ data CategorySite route = CategorySite
       Summary "Get contents of a category"
       :> ErrorResponse 404 "Category not found"
       :> "category"
-      :> Capture "id" (Uid Category)
+      :> Capture "category id" (Uid Category)
```
neongreen: By convention, path variables shouldn't have spaces. You can use `"categoryId"`, `"itemId"` and so on.

neongreen (approved):

PR #255, opened 2019-01-07
Feat/edit category item sections

avele:


front/client/components/Categories.vue
```diff
@@ -73,7 +73,7 @@
           </router-link>
 
           <v-btn
-            class="ma-0 px-1 add-category-btn"
```
Gigiman, line 76: this class was used for test, so lets keep it

front/client/components/Category.vue
```diff
@@ -16,7 +16,7 @@
           <p class="category-top-group"> {{category.group}} </p>
         </div>
         <v-btn
-          class="ml-2 pl-0 add-item-btn"
```
Gigiman, line 19: Class was used for tests, lets keep it

front/client/components/Categories.vue
avele: Test don't work right now anyway and need to be rewritten

PR #256, opened 2019-01-08
Make json fields optional for item info

willbasky:


back/src/Guide/Api/Methods.hs
```diff
@@ -9,6 +9,7 @@ module Guide.Api.Methods where
 
 import Imports
 
+import Control.Monad.Extra (whenJust)
```
neongreen, line 12: 👍

back/src/Guide/Api/Methods.hs
```diff
@@ -121,11 +122,18 @@ setItemInfo :: DB -> RequestDetails -> Uid Item -> CItemInfo -> Guider NoContent
...
   -- TODO diff and merge
-  (editName, _) <- dbUpdate db $ SetItemName itemId $ unH ciiName
-  (editGroup, _) <- dbUpdate db $ SetItemGroup itemId $ unH ciiGroup
-  (editLink, _) <- dbUpdate db $ SetItemLink itemId $ unH ciiLink
-  (editHackage, _) <- dbUpdate db $ SetItemHackage itemId $ unH ciiHackage
-  mapM_ (addEdit db requestDetails) [editName, editGroup, editLink, editHackage]
+  whenJust (unH ciiName) $ \name' -> do
+    (editName, _) <- dbUpdate db $ SetItemName itemId name'
+    addEdit db requestDetails editName
+  when (isJust $ unH ciiGroup) $ do
```
neongreen: Why not `whenJust` here and below?

neongreen: Ah, I see – because those are `Maybe` already.

back/src/Guide/Api/Types.hs
```diff
@@ -433,19 +433,32 @@ instance ToSchema ItemSection where
...
-  , ciiLink    :: Maybe Url  ? "Link to the official site, if exists"
+  { ciiName    :: Maybe Text         ? "Item name"
+  , ciiGroup   :: Maybe (Maybe Text) ? "Item group"
+  , ciiHackage :: Maybe (Maybe Text) ? "Package name on Hackage"
+  , ciiLink    :: Maybe (Maybe Url)  ? "Link to the official site, if exists"
   } deriving (Show, Generic)
 
 instance A.ToJSON CItemInfo where
-  toJSON = A.genericToJSON jsonOptions
+  toJSON cii = A.object
```
neongreen: Almost right.

If the field is `Nothing`, it should be omitted entirely. With this instance, all fields will always be present.

What you want is something like this:

```haskell
object $ catMaybes
  [ ("name" A..=) <$> unH (ciiName cii) 
  , ("group" A..=) <$> unH (ciiGroup cii)
  , ...
  ]
```

PR #258, opened 2019-01-11
Hide db and requestDetails in Guider context

willbasky:


back/src/Guide/Api/Guider.hs
```diff
@@ -13,24 +14,33 @@ import Servant (Handler (..), ServantErr)
...
 
 -- | Custom 'Guider' type holds the 'Config' always on hand.
 newtype Guider a = Guider
-  { runGuider :: ReaderT Config IO a
-  } deriving (Functor, Applicative, Monad, MonadIO, MonadReader Config)
+  { runGuider :: ReaderT ConfigHub IO a
+  } deriving (Functor, Applicative, Monad, MonadIO, MonadReader ConfigHub)
+
+-- | Context of Guider
+data ConfigHub = ConfigHub
```
neongreen: You can just call it `Context`, it's a much more common name.

back/src/Guide/Api/Guider.hs
```diff
@@ -13,24 +14,33 @@ import Servant (Handler (..), ServantErr)
 import Servant.Server.Generic
 
 import Guide.Config (Config)
+import Guide.Api.Utils (RequestDetails)
+import Guide.State (DB)
 
 
 -- | Custom 'Guider' type holds the 'Config' always on hand.
```
neongreen: * It's not just `Config` anymore
* It would be good to mention that it's a `Handler`-like type

```suggestion
-- | A type for Guide handlers. Provides access to everything in 'ConfigHub'.
```

back/src/Guide/Api/Methods.hs
```diff
@@ -247,39 +247,41 @@ dbUpdate db x = liftIO $ do
 -- | Read something from the database.
 dbQuery :: (MonadIO m, EventState event ~ GlobalState, QueryEvent event)
         => DB -> event -> m (EventResult event)
```
neongreen: Perhaps it could be changed to `event -> Guider (EventType event)`? Then we can avoid passing `chDB` everywhere.

willbasky: :+1:

neongreen (approved):
Pushed one more small commit and approved.

PR #259, opened 2019-01-21
Post edit log to matomo

willbasky:


back/guide.cabal
```diff
@@ -50,6 +50,7 @@ library
   exposed-modules:
     Guide.App
     Guide.Api
+      Guide.Api.Matomo
```
neongreen: Matomo isn't a part of our API, so this module should just be `Guide.Matomo`.

back/src/Guide/Api/Matomo.hs
```diff
@@ -0,0 +1,42 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+
+{- | Matomo curl requests -}
```
neongreen: `curl` is a tool for making requests. Calling all HTTP requests "curl requests" is not right.

```haskell
{- | Functions for interacting with Matomo (<https://matomo.org/>, our web analytics).
-}
```

back/src/Guide/Api/Matomo.hs
```diff
@@ -0,0 +1,42 @@
...
+module Guide.Api.Matomo
+       ( Matomo(..)
+       , postMatomo
+       ) where
+
+import Data.IP (IP)
+import Imports
```
neongreen: `import Imports` should come the first

back/src/Guide/Api/Matomo.hs
```diff
@@ -0,0 +1,42 @@
...
+
+-- | Request with log data to send to Matomo
+data Matomo = Matomo
+  { mIP       :: Maybe IP
+  , mUA       :: Maybe Text
+  , mReferrer :: Maybe Text
+  , mTag      :: Edit
+  } deriving (Eq, Show)
+
+-- | Post log of any editions to Matomo.
```
neongreen:
```suggestion
-- | Notify Matomo that an edit has been made.
```

back/src/Guide/Api/Matomo.hs
```diff
@@ -0,0 +1,42 @@
...
+postMatomo :: Matomo -> Guider ()
+postMatomo Matomo{..} = liftIO $ do
+    manager  <- liftIO $ newManager tlsManagerSettings
+    req <- setQueryString
```
neongreen: Please add a comment for each field here.

back/src/Guide/Api/Matomo.hs
```diff
@@ -0,0 +1,42 @@
...
+  { mIP       :: Maybe IP
+  , mUA       :: Maybe Text
+  , mReferrer :: Maybe Text
+  , mTag      :: Edit
+  } deriving (Eq, Show)
+
+-- | Post log of any editions to Matomo.
+postMatomo :: Matomo -> Guider ()
+postMatomo Matomo{..} = liftIO $ do
+    manager  <- liftIO $ newManager tlsManagerSettings
```
neongreen: This `liftIO` is redundant because there's already one.

```suggestion
    manager <- newManager tlsManagerSettings
```

back/src/Guide/Api/Matomo.hs
```diff
@@ -0,0 +1,42 @@
...
+    req <- setQueryString
+      [ ("idsite", Just "1")
+      , ("rec", Just "1")
+      , ("url", toByteString . show <$> mIP)
+      , ("ua", toByteString <$> mUA)
+      , ("urlref", toByteString <$> mReferrer)
+      , ("action_name", Just . toByteString $ show mTag)
```
neongreen: We shouldn't send the whole `Edit` to Matomo, just the tag (i.e. `Edit'BlahBlah`, not `Edit'BlahBlah {foo = blah, bar = blah}`).

back/src/Guide/Api/Matomo.hs
```diff
@@ -0,0 +1,42 @@
...
+      ] <$> parseRequest piwik
+    _ <- httpLbs req manager
+    pure ()
+  where
+    piwik = "POST http://localhost:8080/piwik.php"
```
neongreen: This URL should be configurable in the config.

back/src/Guide/Api/Methods.hs
```diff
@@ -241,6 +242,7 @@ addEdit edit = unless (isVacuousEdit edit) $ do
     Context Config{..} _ RequestDetails{..} <- ask
     dbUpdate $ RegisterEdit edit rdIp time
     dbUpdate $ RegisterAction (Action'Edit edit) rdIp time _baseUrl rdReferer rdUserAgent
+    postMatomo $ Matomo rdIp rdUserAgent rdReferer edit
```
neongreen, line 245: This should be done asynchronously, so that if Matomo is down the wiki will still work.

Also if no Matomo is configured in the config, we shouldn't send analytics at all.

back/src/Guide/Config.hs
```diff
@@ -42,6 +42,8 @@ data Config = Config {
   _portMain      :: Int,          -- ^ Port for the main site.
   _portApi       :: Int,          -- ^ Port for the API.
   _portEkg       :: Int,          -- ^ Port for EKG stats.
+  _portMatomo    :: Maybe Int,    -- ^ Port for Matomo statistic.
+  _pageMatomo    :: Maybe Text,   -- ^ Page on Matomo to send statisic.
```
neongreen: Matomo will not necessarily be running on localhost.

We want just `_matomo :: Maybe Url`, there's no reason to separate the port and the page.

back/src/Guide/Matomo.hs
```diff
@@ -0,0 +1,56 @@
...
+  } deriving (Eq, Show)
+
+-- | Notify Matomo that an edit has been made.
+postMatomo :: Matomo -> Guider ()
+postMatomo Matomo{..} = do
+    Context Config{..} _ _ <- ask
+    case (_portMatomo, _pageMatomo) of
+      (Just port, Just page) -> liftIO $ do
+        flip withAsync (\_ -> pure ()) $ do
+          -- TODO: pushing log to matomo is failed
```
neongreen: It sounds like you're pushing a log to Matomo. I guess you meant "log if the request to Matomo has failed" or something along those lines.

back/src/Guide/Matomo.hs
```diff
@@ -0,0 +1,56 @@
...
+          manager <- newManager tlsManagerSettings
```
neongreen: Creating a new manager is expensive and there's probably a way to avoid it. I think I've seen something related in http-client-tls docs

back/src/Guide/Matomo.hs
```diff
@@ -0,0 +1,56 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+
+{- | Functions for interacting with Matomo (<https://matomo.org/>, our web analytics). -}
```
neongreen: Oh, and can you also add a link to Matomo docs?

https://developer.matomo.org/api-reference/tracking-api

back/src/Guide/Matomo.hs
```diff
@@ -0,0 +1,56 @@
...
+  , mTag      :: Edit
+  } deriving (Eq, Show)
+
+-- | Notify Matomo that an edit has been made.
+postMatomo :: Matomo -> Guider ()
+postMatomo Matomo{..} = do
+    Context Config{..} _ _ <- ask
+    case (_portMatomo, _pageMatomo) of
+      (Just port, Just page) -> liftIO $ do
+        flip withAsync (\_ -> pure ()) $ do
```
neongreen: Why not just `async`?

If I'm reading the docs correctly, `withAsync` will kill the thread then the lambda returns, i. e. immediately.

back/src/Guide/Matomo.hs
```diff
@@ -0,0 +1,65 @@
...
+        manager <- getGlobalManager
+        req <- setQueryString
+          [ ("idsite", Just "1")  -- The ID of the website we're tracking a visit/action for.
+          , ("rec", Just "1")  -- Required for tracking, must be set to one.
+          , ("url", toByteString . show <$> mIP) -- The full URL for the current action.
+          , ("ua", toByteString <$> mUA) -- An override value for the User-Agent HTTP header field.
+          , ("urlref", toByteString <$> mReferrer) -- The full HTTP Referrer URL.
+          , ("action_name", action_name) --  The title of the action being tracked.
+          ] <$> parseRequest (piwik matomo)
+        -- TODO: log if the request to Matomo hass failed
```
neongreen:
```suggestion
        -- TODO: log if the request to Matomo has failed
```

back/src/Guide/Config.hs
```diff
@@ -39,6 +39,7 @@ data Config = Config {
   _adminPassword :: Text,         -- ^ Password for the admin user
   _discussLink   :: Maybe Url,    -- ^ Link to a place to discuss the site.
                                   --    Will be placed in the header
+  _matomoLink    :: Maybe Url,    -- ^ Link of Matomo to send statistic of user`s action. Format of link shoud be like <http://localhost:8081/piwik.php>
```
neongreen: 1. backtick
2. please wrap the comment

neongreen (approved):

PR #260, opened 2019-01-21
uid -> id

vojiranto:


back/src/Guide/Api/Types.hs
```diff
@@ -377,17 +377,17 @@ instance ToSchema CCategoryInfo where
...
+  , cciCreated = H _categoryCreated
+  , cciGroup_  = H _categoryGroup_
   , cciStatus  = _categoryStatus
   }
 
 -- | A "light-weight" client type of 'Category', which gives all available
 -- information about a category
 data CCategoryFull = CCategoryFull
-  { ccfUid         :: Uid Category
+  { ccfId          :: Uid Category
```
willbasky, line 390: Why not `UId -> Id`?

back/src/Guide/Api/Types.hs
```diff
@@ -553,7 +553,7 @@ instance ToSchema CTrait where
 -- | Factory to create a 'CTrait' from a 'Trait'
 toCTrait :: Trait -> CTrait
 toCTrait trait = CTrait
-  { ctUid     = trait ^. uid
+  { ctId     = trait ^. uid
```
willbasky, line 556: `uid` can`t change to `id` but maybe it can change to `tId` or something better.

PR #263, opened 2019-01-24
Delete/item/ id   escape should work

Gigiman:
added close add-category dialog when esc button is pressed


front/config.json
```diff
@@ -1,3 +1,3 @@
 {
```
avele: Remove changes in config

Gigiman: done

PR #264, opened 2019-01-25
Task/log monad

vojiranto:


back/src/Guide/Config.hs
```diff
@@ -78,6 +84,9 @@ instance FromJSON Config where
     _portEkg       <- o .:? "port-ekg"       .!= _portEkg def
     _cors          <- o .:? "cors"           .!= _cors def
     _ekg           <- o .:? "ekg"            .!= _ekg def
+    _logToStderr   <- o .:? "log-to-stderr"  .!= _logToStderr def
+    _logToFile     <- o .:? "log-to-file"    .!= _logToFile def
+    _logTimeFormat <- o .:? "log-time-format" .!= _logTimeFormat def
```
willbasky, line 110: What about make `.!=` in row?

back/src/Guide/Api/Methods.hs
```diff
@@ -36,18 +37,23 @@ import qualified Guide.Search as Search
...
-getCategories = dbQuery GetCategories <&> map toCCategoryInfo
+getCategories = do
+  debugT "getCategories"
+  dbQuery GetCategories <&> map toCCategoryInfo
 
 -- | Get a single category and all of its items.
 getCategory :: Uid Category -> Guider CCategoryFull
-getCategory catId = toCCategoryFull <$> getCategoryOrFail catId
+getCategory catId = do
+  debugT $ "getCategory: " +| catId |+ ""
```
willbasky: `+|` - fmt?

back/src/Guide/Api/Methods.hs
```diff
@@ -217,6 +253,7 @@ moveTrait itemId traitId CMove{..} = do
 -- Returns at most 100 results.
 search :: Text -> Guider [CSearchResult]
 search searchQuery = do
+  debugT $ "search: " +|| searchQuery ||+ " "
```
neongreen: Use `push` here as well?

back/src/Guide/Api/Methods.hs
```diff
@@ -225,24 +262,29 @@ search searchQuery = do
...
         => event -> Guider (EventResult event)
 dbQuery x = do
+  debugT $ "dbQuery: " +|| x ||+ ""
   Context{..} <- ask
   liftIO $ Acid.query cDB x
 
 -- Call this whenever any user-made change is applied to the database.
 addEdit :: Edit -> Guider ()
-addEdit edit = unless (isVacuousEdit edit) $ do
+addEdit edit = do
```
neongreen: And here?

back/src/Guide/Logger/Init.hs
```diff
@@ -0,0 +1,68 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving  #-}
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
neongreen: When there are orphans, let's also always add a comment like this so that it would be easier to check if the pragma is outdated or not:

```haskell
{-# OPTIONS_GHC -fno-warn-orphans #-} -- for "instance Read Df1.Level"
```

back/src/Guide/Logger/Init.hs
```diff
@@ -0,0 +1,68 @@
...
+        logLvlMark =
+            timeMark <>
+            (printPath $ toList path) <>
+            T.pack (show lvl)
+
+      when _logToStderr   $ sayErr formattedMsg
+      whenJust _logToFile $ \fileName -> do
+        T.appendFile fileName (formattedMsg <> "\n")
+
+unPath :: Df1.Path -> Text
```
neongreen: Please add a haddock with examples of how different `Path`s are shown.

back/src/Guide/Logger/Methods.hs
```diff
@@ -0,0 +1,41 @@
...
+import Imports
+
+import qualified Df1
+import qualified Di.Core as DC
+import Di.Monad
+import qualified Di
+import Guide.Logger.Types
+import qualified Data.Text.Lazy as L
+
+value :: Show a => a -> Di.Value
```
neongreen: Haddock please

back/src/Guide/Logger/Types.hs
```diff
@@ -0,0 +1,9 @@
+module Guide.Logger.Types where
+
+import Imports
+import Di.Core
+import Di.Monad
+import Df1
+
+type DefDiT = DiT Level Path Message IO
```
neongreen: Haddocks please

back/src/Guide/Matomo.hs
```diff
@@ -36,6 +37,7 @@ data Matomo = Matomo
 -- | Notify Matomo that an edit has been made.
 postMatomo :: Matomo -> Guider ()
 postMatomo Matomo{..} = do
+    debugT "postMatomo"
```
neongreen: `push`?

back/src/Imports.hs
```diff
@@ -10,6 +10,8 @@ module Imports
   module X,
   LByteString,
   LText,
+  TL.toStrict,
+  TL.fromStrict
```
neongreen: In the `To` module we have `toText` and `toLText`; those should be used instead.

back/tests/Api.hs
```diff
@@ -0,0 +1,300 @@
...
+import Guide.Api.Types
+import Guide.Types.Core
+import Guide.Utils (Uid (..))
+import qualified Test.Hspec as H
+
+tests :: H.SpecWith ()
+tests = H.describe "api" $ do
+  H.it "fail request" $ do
+    request <- makeRequest
+      (Host "http://localhost/fail")
```
neongreen: It is kinda confusing because without looking at `makeRequest` it is hard to guess that `makeRequest` will sneakily add `:4400` to the host. It's better to just have something like `makeRequest "GET" "/fail"` and prepend `localhost:4400` in `makeRequest` itself.

back/tests/Api.hs
```diff
@@ -0,0 +1,300 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+
+module Api where
```
neongreen: Haddock please (what is this module testing?)

back/tests/WebSpec.hs
```diff
@@ -25,23 +26,92 @@ import Control.Monad.Catch
...
+-- Spec
+import qualified Test.Hspec as H
+import System.IO
 
 -----------------------------------------------------------------------------
 -- Tests
 -----------------------------------------------------------------------------
 
+logFile :: FilePath
+logFile = "/tmp/test_guide.log"
```
neongreen: This only works on Unix systems. It's better to use `withTempFile` or something, see http://hackage.haskell.org/package/temporary-1.3/docs/System-IO-Temp.html#v:withTempFile

back/tests/WebSpec.hs
```diff
@@ -25,23 +26,92 @@ import Control.Monad.Catch
...
+    hClose logFileHandle
+    writeFile logFile ""
+    pure logs
+  H.describe "Logging of init" $ do
+    H.it "spock is runned" $ [re|Spock is running on port|] `isIn` logs
+    H.it "api is runned" $ [re|API is running on port|] `isIn` logs
+  H.describe "Logging of api" $ do
+    H.describe "Categories" $ do
+      H.it "get categories request" $ [re|getCategories|] `isIn` logs
+      H.it "create ctegory request" $ [re|createCategory|] `isIn` logs
```
neongreen:
```suggestion
      H.it "create category request" $ [re|createCategory|] `isIn` logs
```

back/tests/WebSpec.hs
```diff
@@ -25,23 +26,92 @@ import Control.Monad.Catch
...
+logTest :: Spec
+logTest = H.describe "test of logger" $ do
+  logs <- H.runIO $ do
+    logFileHandle <- openFile logFile ReadWriteMode
+    logs <- getLines logFileHandle
+    hClose logFileHandle
+    writeFile logFile ""
+    pure logs
+  H.describe "Logging of init" $ do
+    H.it "spock is runned" $ [re|Spock is running on port|] `isIn` logs
```
neongreen: "Spock is run" (run ran run)

neongreen: Or even better "Spock init message is present" or something

back/tests/WebSpec.hs
```diff
@@ -25,23 +26,92 @@ import Control.Monad.Catch
...
+    H.it "api is runned" $ [re|API is running on port|] `isIn` logs
+  H.describe "Logging of api" $ do
+    H.describe "Categories" $ do
+      H.it "get categories request" $ [re|getCategories|] `isIn` logs
+      H.it "create ctegory request" $ [re|createCategory|] `isIn` logs
+      H.it "get category by id" $ [re|getCategory|] `isIn` logs
+      H.it "delete category by id request" $ [re|deleteCategory|] `isIn` logs
+      H.it "modify notes to category request" $ [re|setCategoryNotes|] `isIn` logs
+      H.it "modify info to category request" $ [re|setCategoryInfo|] `isIn` logs
+    H.describe "Item" $ do
```
neongreen: I feel it's not necessary to test logging for every single endpoint.

If somebody is careless enough that they see 20 endpoints with logging and still forget to add logging to the new endpoint they're writing, they will also sure as heck forget to add a test for it.

back/tests/WebSpec.hs
```diff
@@ -25,23 +26,92 @@ import Control.Monad.Catch
...
+      H.it "set item ecosystem" $ [re|setItemEcosystem|] `isIn` logs
+      H.it "set item notes" $ [re|setItemNotes|] `isIn` logs
+    H.describe "Trait" $ do
+      H.it "create new trait" $  [re|createTrait|] `isIn` logs
+      H.it "delete trait" $ [re|deleteTrait|] `isIn` logs
+      H.it "get trait" $ [re|getTrait|] `isIn` logs
+      H.it "modify trait" $ [re|setTrait|] `isIn` logs
+      H.it "move trait" $ [re|moveTrait|] `isIn` logs
+  
+    H.describe "Errors (exceptions)" $ do
```
neongreen: Same goes here.

* If we want to test that exceptions are logged (in general), testing three endpoints is unnecessary. It's better to have a single test where you throw a specific exception and see if your logging setup catches it.

* If you're testing _logic_ this way, then it absolutely should be done by looking at HTTP responses, not at lines in the log.

back/tests/WebSpec.hs
```diff
@@ -25,23 +26,92 @@ import Control.Monad.Catch
...
+        [re|ServantErr {errHTTPCode = 404, errReasonPhrase = "Category not found"|] `isIn` logs
+      H.it "Merge conflict occurred" $
+        [re|ServantErr {errHTTPCode = 409, errReasonPhrase = "Merge conflict occurred"|] `isIn` logs
+      H.it "Item not found" $
+        [re|ServantErr {errHTTPCode = 404, errReasonPhrase = "Item not found"|] `isIn` logs
+
+
+
+isIn :: RE -> String -> H.Expectation
+isIn  reg text = (`H.shouldBe` True) $ matched $ text ?=~ reg
```
neongreen: It's better to write something with `H.expectationFailure` because then you'll get the actual regex that failed, and not "expected True, got False".

You should also add `HasCallStack =>` to get stack traces.

back/tests/Api.hs
```diff
@@ -0,0 +1,300 @@
...
+runRequestNoBody, runFailRequest :: Request -> IO Status
+runRequestNoBody request = getResponseStatus <$> httpNoBody request
+runFailRequest           = runRequestNoBody
+
+runRequest :: Yaml.FromJSON a => Request -> IO (Status, a)
+runRequest request = do
+  response <- httpJSON request
+  pure (getResponseStatus response, getResponseBody response)
+
+newtype Host   = Host String
```
neongreen: `localhost:4400` is host, but `localhost/fail` is a URL

back/tests/WebSpec.hs
willbasky: So, either will we stop to add new tests or will we delete several to be not the whole?

back/tests/WebSpec.hs
willbasky: @vojiranto what purposes did you have for the piece of code?

back/tests/WebSpec.hs
neongreen: If they are testing only logging, we can delete all of them (well, leave one).

back/tests/WebSpec.hs
neongreen: We can leave one of these, too. 

It checks that Servant exceptions are logged.

back/src/Guide/Api/Server.hs
```diff
@@ -74,29 +73,33 @@ type FullApi =
   Api :<|>
   SwaggerSchemaUI "api" "swagger.json"
 
-fullServer :: DB -> Config -> Server FullApi
-fullServer db config =
-  api db config :<|>
+fullServer :: DB -> DefDi -> Config -> Server FullApi
```
neongreen: Haddock.

back/tests/WebSpec.hs
```diff
@@ -18,30 +19,79 @@ import Data.Text (Text)
...
 -----------------------------------------------------------------------------
 
 tests :: IO ()
-tests = run $ do
-  mainPageTests
-  categoryTests
-  itemTests
-  markdownTests
+tests = withSystemTempFile "test_guide.log" $ \logFile logFileHandle -> do
+  hClose logFileHandle -- It closed for api tests were able to append logs.
```
neongreen: Not very comprehensible

back/tests/WebSpec.hs
```diff
@@ -18,30 +19,79 @@ import Data.Text (Text)
...
+  run logFile $ do
+    mainPageTests
+    categoryTests
+    itemTests
+    markdownTests
+    Api.tests
+  hspec $ logTest logFile
+
+getLines :: Handle -> IO String
```
neongreen: What does this do?

Why does it do it in such a way?

back/tests/WebSpec.hs
```diff
@@ -589,8 +639,8 @@ getCurrentRelativeURL = do
       maybe "" uriRegName (uriAuthority u) `shouldBe` "localhost"
       return u
 
-run :: Spec -> IO ()
-run ts = do
+run :: FilePath ->  Spec -> IO ()
```
neongreen: Haddock

neongreen:
@willbasky this can be re-reviewed.

Since I pushed some unrelated changes here as well, here are descriptions for the new commits (except for ones that only add comments):

* https://github.com/aelve/guide/pull/264/commits/503907fae86d3700c5af0370a69851ccbaa12af2 -- moves error logging into `instance MonadError Guider` and adds a logger for uncaught exceptions.

* https://github.com/aelve/guide/pull/264/commits/6fbeab91736b1ce3739d847f32b437c87e81cc6a -- removes `.ghci` because it prevents the project from being loaded in GHCi.

* https://github.com/aelve/guide/pull/264/commits/fac5ef0ea245e7e21e73d918c74c3c71b4536475, https://github.com/aelve/guide/pull/264/commits/c0614a836651932be5e17ffed2d8fbd0d57d39ff -- changes log formatting (though we'll have to change it further, I looked at a big log and it's not pretty at all)

* https://github.com/aelve/guide/pull/264/commits/960baedab527fc68f442c86acee91beaea31196d -- removes log levels like `emergency`, `alert`, etc because they are useless for us.

* https://github.com/aelve/guide/pull/264/commits/28f1bb7c5f6340c96618cd68d55f0db81e4f58bb, https://github.com/aelve/guide/pull/264/commits/b5b5b884a7a8ca62c5fe8e956df4509d894b3272 --  renames `Guide.Logger.{Init,Methods}` to `Guide.Logger.{Run,Functions}`, renames `debugT` and so on to more standard `logDebug`, and introduces name like `Logger` and `LoggerT` because `Di` and `DiT` are rather bad names.

* https://github.com/aelve/guide/pull/264/commits/654d49f2162a926d5937f61bb160574416c37470 -- adds more logging (except for `checkConflict` because it seems to not be useful there, but maybe I'm wrong).

* https://github.com/aelve/guide/pull/264/commits/1550ad956228ea830f5fbd664f974ae7ee46bbab -- removes `GuiderServer` because it was an extra name that wasn't contributing anything.

* https://github.com/aelve/guide/pull/264/commits/171e10545ecdff10813141fd5db41fa3df6b11e2 -- renames the `Api` module to `ApiSpec` to be consistent with other modules.

* https://github.com/aelve/guide/pull/264/commits/416cad8603d6441b71bd81907207ec776ab8d957 -- removes the footer test because it breaks all the time and is not that useful.

.travis.yml
```diff
@@ -49,6 +49,16 @@ jobs:
         - docker push "quay.io/aelve/guide:$BRANCH--back"
       if: type = push
 
+    - stage: "Test haddocked <= 251"
```
neongreen: This is not easy to understand for someone who has no idea what this code is supposed to do. It's nicer to spell out the error message, e.g. "FAIL: the limit is 251 undocumented top-level definitions, but $notHaddockedAfter were found".

It would also be nice to add a comment before `- stage: Test haddocked <= 251`, saying that we're trying to get the number of undocumented definitions to zero.

"haddocked" sounds pretty weird, though technically it's correct. I would recommend "documented" (or "undocumented" in this case).

back/tests/WebSpec.hs
```diff
@@ -48,9 +46,6 @@ tests = withSystemTempFile "test_guide.log" $ \logFile logFileHandle -> do
     itemTests
     markdownTests
     ApiSpec.tests
-  -- TODO: get rid of the 'threadDelay' by using 'withAsync' instead of
-  -- 'Slave.fork'
-  threadDelay 1000000
```
neongreen: Does this actually work? Because in other places we still have `Slave.fork` and I'm not sure if it mixes with `withAsync`. If it does, great.
There are just three ones at Main module.

willbasky: Tests passed. A site server works. Is it actually? Likely, it is.
What about to swap `Slave.fork` with `withAsync`?

back/tests/WebSpec.hs
willbasky: It seems the quoted piece is outdated. Actual version is 
```
  -- Close the log file because otherwise 'run' won't be able to open it
  hClose logFileHandle
```

back/tests/WebSpec.hs
willbasky: The quoted piece is outdated also. There is no such function already.

PR #268, opened 2019-02-27
Category info edit

Gigiman:
Editing category info functionality added


front/client/components/CategoryInfoEdit.vue
```diff
@@ -0,0 +1,177 @@
...
+    if (status === 'CategoryStub') {
+      return 'Stub'
+    } else if (status === 'CategoryWIP') {
+      return 'Work in progress'
+    } else if (status === 'CategoryFinished') {
+      return 'Complete'
+    }
+  }
+
+  setCategoryStatusRaw () {
```
avele: 1) This function doesnt set anything, it returns value. So name is really confusing.
2) Second, have you heard about switch? 
3) There are CategoryStatus enum, check it and replace raw strings to its properties.

front/client/components/CategoryInfoEdit.vue
```diff
@@ -0,0 +1,177 @@
...
+  onEcosysemChange () {
+    this.onCheckboxChange(this.ecosystem, 'ItemEcosystemSection')
+  }
+
+  @Watch('notes')
+  onNotesChange () {
+    this.onCheckboxChange(this.notes, 'ItemNotesSection')
+  }
+
+  transformCategoryStatus (status: string) {
```
avele, line 121: Next function is almost the same.

front/client/components/CategoryInfoEdit.vue
```diff
@@ -0,0 +1,177 @@
...
+
+  itemValidationRules: Function[] = [
+    (x: string) => !!x || 'Item name can not be empty'
+  ]
+
+  beforeMount () {
+    this.$vuetify.icons.checkboxOff = this.$vuetify.icons.square
+  }
+
+  onCheckboxChange(checkboxType: boolean, checkboxName: string) {
```
avele: This function is hard to understand. Refactor it. You can use array as v-model for all checkboxes so you dont need all this watchers and this function too. Check out vue doc, or event vuetify doc (there is such example with vuetify checkboxes)
Also, there are native array functions which do the same as "_includes" and "_remove", so you dont need to import excess code from lodash.

front/client/components/CategoryInfoEdit.vue
```diff
@@ -0,0 +1,177 @@
...
+  isValid: boolean = false
+
+  items = ['Complete', 'Work in progress', 'Stub']
+
+  itemValidationRules: Function[] = [
+    (x: string) => !!x || 'Item name can not be empty'
+  ]
+
+  beforeMount () {
+    this.$vuetify.icons.checkboxOff = this.$vuetify.icons.square
```
avele, line 109: Wtf is this? Why you replace one icon by another? And what happens if I want to use "checkboxOff" icon somewhere after this?

front/client/components/CategoryInfoEdit.vue
```diff
@@ -0,0 +1,177 @@
...
+          flat
+          color="primary"
+          @click.native="close"
+        >
+          Cancel
+        </v-btn>
+      </v-card-actions>
+    </v-card>
+  </v-dialog>
+</template> -->
```
avele: Some strange arrrow down here.

front/client/page/SearchResults.vue
```diff
@@ -17,7 +17,7 @@
             »
           <a-link
             openInNewTab
```
avele, line 19: I think its better to remove all this chages and replace links to router-link pointing to category component in another pull-request

front/client/components/CategoryInfoEdit.vue
```diff
@@ -0,0 +1,177 @@
...
+  prosCons: boolean = true
+  ecosystem: boolean = true
+  notes: boolean = true
+  sections: any[] = []
+  isValid: boolean = false
+
+  items = ['Complete', 'Work in progress', 'Stub']
+
+  itemValidationRules: Function[] = [
+    (x: string) => !!x || 'Item name can not be empty'
```
avele: You call both 'Title' and 'Group'. Make accurate validation rules for each input (you can even do it in markup)

front/client/components/CategoryInfoEdit.vue
Gigiman: done

front/client/components/CategoryInfoEdit.vue
Gigiman: this is the problem with the svg icons we were discussing, some icons cant be replaced with the svg icon. Icon replacement is a hack from vuetify issue

front/client/components/CategoryInfoEdit.vue
Gigiman: done

front/client/components/CategoryInfoEdit.vue
Gigiman: done

front/client/components/CategoryInfoEdit.vue
Gigiman: done

front/client/components/CategoryInfoEdit.vue
Gigiman: refactored

front/client/page/SearchResults.vue
Gigiman: added to task list in notion

PR #270, opened 2019-03-07
Fix json instances for toc

willbasky:


back/src/Guide/Api/Types.hs
```diff
@@ -644,6 +642,42 @@ toCHeading h = CHeading
   , chSlug    = H $ headingSlug h
   }
 
+-- | Custom 'CForest' is the same as 'Forest'.
+-- | It needed to customize JSON instances for TOC to be list of object, not list of lists.
```
neongreen: "list of objects"

back/src/Guide/Api/Types.hs
```diff
@@ -644,6 +642,42 @@ toCHeading h = CHeading
...
+type CForest a = [CTree a]
+
+-- | Custom 'CTree' is the same as 'Tree'.
+-- | It needed to customize JSON instances for TOC to be list of object, not list of lists.
+data CTree a = CNode
+  { cRootLabel :: a
```
neongreen: Since the type is called "CTree", the prefix is "ct", not "c"

back/src/Guide/Api/Types.hs
```diff
@@ -644,6 +642,42 @@ toCHeading h = CHeading
...
+  , cSubForest :: CForest a
+  } deriving (Show, Generic)
+
+instance A.ToJSON (CTree CHeading) where
+  toJSON (CNode CHeading{..} cSubForest) = A.object
+    [ "content" A..= chContent
+    , "slug"    A..= chSlug
+    , "forest"  A..= cSubForest
```
neongreen: Please ask @avele or @Gigiman about the name, `"forest"` seems like not the best name.

back/src/Guide/Api/Types.hs
```diff
@@ -644,6 +642,42 @@ toCHeading h = CHeading
...
+instance A.FromJSON (CTree CHeading) where
+  parseJSON = A.withObject "CTree" $ \o -> do
+    chContent  <- o A..: "content"
+    chSlug     <- o A..: "slug"
+    cSubForest <- o A..: "forest"
+    let cRootLabel = CHeading{..}
+    pure CNode{..}
+
+instance ToSchema a => ToSchema (CTree a) where
+  declareNamedSchema = genericDeclareNamedSchema schemaOptions
```
neongreen: Pretty sure this will not give you the right Swagger model.

I would recommend changing the types to something like this:

```haskell
data CTocHeading = CTocHeading
  { cthContent     :: CMarkdown
  , cthSlug        :: Text           ? "In-page anchor for linking"
  , cthSubheadings :: [CTocHeading]
  }
```

neongreen: (And then the generic model will be the right one.)

back/src/Guide/Api/Types.hs
```diff
@@ -644,6 +642,42 @@ toCHeading h = CHeading
...
+
+-- | 'toCTree' used to make type with customizable JSON instances.
```
neongreen: This comment is repeated three times in the code.

You can do something like this (or whatever else):

```suggestion
-- | 'toCTree' converts a table of contents into the format expected by the frontend.
```

neongreen (approved):

PR #274, opened 2019-03-27
Rss icon and link added

Gigiman:
Rss icon and link added, category info moved to the separate component

### DIFF
--- a/front/client/components/Category.vue
+++ b/front/client/components/Category.vue
@@ -8,7 +8,7 @@
           :categoryTitle="category.title"
           :categoryGroup="category.group"
           :categoryUrl="categoryUrl"
-          @openDialog="openAddItemDialog"
+          @openAddItemDialog="openAddItemDialog"
         />
       </template>
 
--- a/front/client/components/Category.vue
+++ b/front/client/components/Category.vue
@@ -6,6 +6,12 @@
           v-if="category"
           class="category-top-data"
         >
+          <a-link
+            openInNewTab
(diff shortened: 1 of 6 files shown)


front/client/components/Category.vue
```diff
@@ -1,45 +1,16 @@
...
-        </v-btn>
-      </div>
+      <template v-if="category">
+        <category-info
+          :category="category" 
+          :categoryId="categoryId"
+          :categoryTitle="category.title"
+          :categoryGroup="category.group"
+          :categoryUrl="categoryUrl"
+          @openDialog="openAddItemDialog"
```
avele: Rename "openDialog" to something more specific, openAddItemDialog - is ok

front/client/components/CategoryItemTraits.vue
```diff
@@ -28,7 +28,7 @@
           <category-item-btn
             slot="activator"
             size="22px"
-            :iconSize="14"
+            :iconSize="`14`"
```
avele: You dont need to wrap 14 in \` \` if you want it to be passed as string. Just remove `:` and it will be passed to component as string. Like this `iconSize="14"`

front/client/components/Category.vue
Gigiman: done

front/client/components/CategoryItemTraits.vue
Gigiman: done

avele (approved):

PR #282, opened 2019-04-26
Fix dockerfile and readme

willbasky:


docker/front/Dockerfile
```diff
@@ -6,6 +6,10 @@ ENV LC_ALL=C.UTF-8
 # NODE_ENV used to run front server on production
 ENV NODE_ENV=production
 
+# These variables are for independent from front/config.js
```
neongreen: Weird English

front/README.md
```diff
@@ -1,15 +1,21 @@
...
- - `cd dist`
- - `npm i`
- - `node server`
+  Requirements: Node version should be >= 11, because of using fs async methods which experimental now.
+
+1. `git pull`
+2. `npm i`
+
+3. Now you have a choice how to set PORT and API_URL.
+   First, set them in front/config.js once.
```
neongreen: > Either set them in `front/config.js`, or use export ...

docker/front/Dockerfile
neongreen: Do these variables override ones in front/config.js, or do they serve as defaults when front/config.js doesn't have any variables set?

willbasky: Both.

neongreen: If front/config.js says that port is 4444 and the shell variable says PORT=5555, which value will actually be used?

willbasky: 5555

neongreen:
```suggestion
# Overrides for variables in front/config.js
```

neongreen (approved):

PR #290, opened 2019-05-17
Fix some bugs in the Travis config and reduce the number of jobs

neongreen:


.travis.yml
```diff
@@ -29,50 +29,49 @@ cache:
...
         - cd front
         - npm install
         - npm run build
         - cd -
-      script:
-        - export BRANCH="$(echo "$TRAVIS_BRANCH" | tr '/' '-')"
-        - make front/travis-docker "tag=$BRANCH--front"
-        - docker login quay.io -u "$DOCKER_USER" -p "$DOCKER_PASS"
-        - docker push "quay.io/aelve/guide:$BRANCH--front"
-      if: type = push
```
neongreen, line 43: I changed this to a bash `if` to make it more clear that the stage should run for both PRs and pushes, and only the Docker part should run for pushes.

.travis.yml
```diff
@@ -29,50 +29,49 @@ cache:
...
+        - |
+          if [ "$TRAVIS_EVENT_TYPE" = "push" ]; then
+            export BRANCH="$(echo "$TRAVIS_BRANCH" | tr '/' '-')"
+            make front/travis-docker "tag=$BRANCH--front"
+            docker login quay.io -u "$DOCKER_USER" -p "$DOCKER_PASS"
+            docker push "quay.io/aelve/guide:$BRANCH--front"
+          fi
 
-    - stage: "Build backend dependencies"
```
neongreen, line 45: I consolidated the stages because each stage brought about 3 minutes of CI time overhead (loading the cache, unloading the cache).

.travis.yml
```diff
@@ -29,50 +29,49 @@ cache:
...
       script:
         - stack haddock --fast --no-haddock-deps --no-terminal --haddock-arguments='--no-warnings' 2> haddock.log
-        - awk '/\(src\//' haddock.log | awk 'END { print NR }' > unDocumentedAfter
-        - echo "unDocumentedBefore == 251"
-        - awk '{ print "unDocumentedAfter == " $1 }' unDocumentedAfter
-        - bash -c 'if [[ 251 < `cat unDocumentedAfter` ]]; then >&2 echo "FAIL. The current limit is 251 of undocumented top-level definitions, but unDocumentedAfter is out of the limit. Please, add comment to top-level undocumented definitions!"; exit 1; else echo "OK. unDocumentedBefore >= unDocumentedAfter. Yes, it is convenience to have documented code!"; fi'
+        - awk '/\(src\//' haddock.log | awk 'END { print NR }' > undocumented
+        - awk '{ print "undocumented == " $1 }' undocumented
+        - |
+          if [[ `cat undocumented` -gt 251 ]]; then
```
neongreen, line 71: Rewrote from `bash -c` to an easier-to-follow version.

neongreen: Also, using `<` was incorrect because it compared strings, not numbers.

neongreen:
This PR brings CI time from 40 minutes to 30 minutes.

The absolute possible minimum is something like 10–15 minutes, I think, but it would require moving from Travis to Gitlab (Travis caching is slow).

.travis.yml
willbasky: :+1:

willbasky (approved):

PR #291, opened 2019-05-18
Further Travis improvements

neongreen:


neongreen:
This gets us to 25 minutes, but we should be able to get 20.

.travis.yml
```diff
@@ -4,15 +4,8 @@ dist: trusty
 language: generic
 
 addons:
-  apt:
```
neongreen, line 7: Moved to individual stages.

.travis.yml
```diff
@@ -45,37 +40,46 @@ jobs:
 
     - stage: "Build the backend and upload a Docker image"
       before_script:
+        - sudo apt-get install -y libgmp-dev
         - curl -sSL https://get.haskellstack.org/ | sh
+        - stack setup
```
neongreen, line 45: Installs GHC (same as before but it's nicer to do it as a separate step and get Travis timings for this step)

.travis.yml
```diff
@@ -45,37 +40,46 @@ jobs:
...
+        # Upload the Docker image
         - |
           if [ "$TRAVIS_EVENT_TYPE" = "push" ]; then
             export BRANCH="$(echo "$TRAVIS_BRANCH" | tr '/' '-')"
             make back/travis-docker "tag=$BRANCH--back"
             docker login quay.io -u "$DOCKER_USER" -p "$DOCKER_PASS"
             docker push "quay.io/aelve/guide:$BRANCH--back"
           fi
-
-    # Our goal is full documented code with ZERO undocumented definitions.
```
neongreen, line 60: I consolidated this step and the previous step.

neongreen:
This PR brings us down to 23–25 minutes.

.travis.yml
```diff
@@ -4,15 +4,8 @@ dist: trusty
...
 
 addons:
-  apt:
-    packages:
-    # GHC
-    - libgmp-dev
-    # Selenium
-    - xvfb
-    # Testcafe
-    - fluxbox
```
willbasky, line 14: Don't we need these packages anymore?

neongreen: We do, but now we install them only when we need them:

image

willbasky (approved):

PR #328, opened 2019-07-09
Create tables with hasql

willbasky:


back/guide.cabal
```diff
@@ -51,6 +51,7 @@ library
     Guide.App
     Guide.Api
       Guide.Api.Methods
+      Guide.Api.Migration
```
neongreen: This module is not related to the API, so a better name would be appropriate. E.g. `Guide.Db.Schema` (since there are no migrations here it should hardly be called "migration").

back/src/Guide/Api/Migration.hs
```diff
@@ -0,0 +1,131 @@
...
+
+main :: IO ()
+main = do
+  conn <- connection
+  dbCreate conn
+
+connection :: IO Connection
+connection = do
+  eiConnection <- HC.acquire connectionSettings
+  let conn = either (error . show) id eiConnection
```
neongreen: I would suggest:

```haskell
connection =
  either (error . show) pure =<< 
  HC.acquire connectionSettings
```

Why? Because this way the `error` will be thrown even if the connection is not used, instead of being thrown lazily.

back/src/Guide/Api/Migration.hs
```diff
@@ -0,0 +1,131 @@
...
+    [createTypeProcons, createDbCategories, createDbItems, createDbTraits, createDbUsers, createDbEdits]
+  mapM_ (either print pure) result
+
+createTypeProcons :: Session ()
+createTypeProcons = HS.sql "CREATE TYPE trait_type AS ENUM ('pro', 'con');"
+
+createDbTraits :: Session ()
+createDbTraits = HS.sql $ B.intercalate " "
+  [ "CREATE TABLE traits ("
+  ,   "uid text PRIMARY KEY,"           -- Unique trait ID
```
neongreen: You can use http://hackage.haskell.org/package/neat-interpolation-0.3.2.4/docs/NeatInterpolation.html here together with some helper function like `sqlText`. Postgres supports Haskell-style comments so you don't need to take special care of them.

back/src/Guide/Api/Migration.hs
```diff
@@ -0,0 +1,131 @@
...
+  ,     "NOT NULL"
+  , ");"
+  ]
+
+createDbEdits :: Session ()
+createDbEdits = HS.sql $ B.intercalate " "
+  [ "CREATE TABLE pending_edits ("
+  ,   "uid bigserial PRIMARY KEY,"   -- Unique id
+  ,   "edit json NOT NULL,"          -- Edit in JSON format
+  ,   "ip cidr,"                     -- IP address of edit maker
```
neongreen:
```suggestion
  ,   "ip inet,"                     -- IP address of edit maker
```

back/src/Guide/Api/Migration.hs
```diff
@@ -0,0 +1,131 @@
...
+  pure conn
+
+connectionSettings :: Settings
+connectionSettings = HC.settings "localhost" 5432 "postgres" "3" "guide"
+
+dbCreate :: Connection -> IO ()
+dbCreate conn = do
+  result <- mapM (\s -> HS.run s conn)
+    [createTypeProcons, createDbCategories, createDbItems, createDbTraits, createDbUsers, createDbEdits]
+  mapM_ (either print pure) result
```
neongreen: Errors should be printed to stderr, not stdout. In addition, we should crash if any errors happened. So, you should collect all errors with `lefts` and crash if the list is non-empty.

back/src/Guide/Api/Migration.hs
```diff
@@ -0,0 +1,131 @@
...
+
+createTypeProcons :: Session ()
```
neongreen: createTypeTraitType

back/src/Guide/Api/Migration.hs
```diff
@@ -0,0 +1,131 @@
...
+createTypeProcons = HS.sql "CREATE TYPE trait_type AS ENUM ('pro', 'con');"
+
+createDbTraits :: Session ()
```
neongreen: Perhaps `createTableTraits` is better

back/src/Guide/Db/Schema.hs
```diff
@@ -0,0 +1,146 @@
...
+connectionSettings :: Settings
+connectionSettings = HC.settings "localhost" 5432 "postgres" pass "guide"
+
+pass :: ByteString
+pass = ""
+
+dbCreate :: Connection -> IO ()
+dbCreate conn = do
+  result <- mapM (\s -> HS.run s conn) sessionList
+  mapM_ (hPrint stderr) $ lefts result
```
neongreen: Well, you still have to fail though

back/src/Guide/Db/Schema.hs
```diff
@@ -0,0 +1,146 @@
...
+  , createDbItems
+  , createDbTraits
+  , createDbUsers
+  , createDbEdits
+  ]
+
+createTypeTraitType :: Session ()
+createTypeTraitType = HS.sql "CREATE TYPE trait_type AS ENUM ('pro', 'con');"
+
+createDbTraits :: Session ()
```
neongreen: If you decide not to rename the function, can you provide reasoning for why "db" is better? My reasoning for why "table" is better is that Postgres has [`CREATE DATABASE`](https://www.postgresql.org/docs/9.0/sql-createdatabase.html) as well and it's a different thing from `CREATE TABLE`.

willbasky: ops, I missed that comment.

back/src/Guide/Db/Schema.hs
```diff
@@ -0,0 +1,150 @@
...
+sessionList =
+  [ createTypeProCons
+  , createTableCategories
+  , createTableItems
+  , createTableTraits
+  , createTableUsers
+  , createTableEdits
+  ]
+
+createTypeProCons :: Session ()
```
neongreen: `ProCon`

neongreen:
Ok, please document all definitions and this can be merged.

back/src/Guide/Db/Schema.hs
```diff
@@ -46,39 +76,70 @@ dbPass = "3"
...
+-- | Like 'HS.run', but errors out in case of failure.
+run' :: Session a -> Connection -> IO a
+run' s c = either (error . show) pure =<< HS.run s c
+
+----------------------------------------------------------------------------
+-- Schema version table
+----------------------------------------------------------------------------
+
+-- | Get schema version (i.e. the version of the last migration that was
+-- run). If the @schema_version@ table doesn't exist, creates it.
```
willbasky: First, I saw `create`, command to user to create a table, later i saw `createS`.

back/src/Guide/Db/Schema.hs
```diff
@@ -2,33 +2,63 @@
...
+-- Does not create the @guide@ database if it does not exist yet. Use
+-- @CREATE DATABASE@ when running for the first time.
+--
+-- TODO: check schema hash as well, not just schema version?
+setupDatabase :: IO ()
+setupDatabase = do
+  conn <- connect
+  mbSchemaVersion <- run' getSchemaVersion conn
+  case mbSchemaVersion of
+    Nothing -> formatLn "No schema found, running all migrations"
```
willbasky: > running all migrations

what does it means?

back/src/Guide/Db/Schema.hs
```diff
@@ -2,33 +2,63 @@
...
-connection :: IO Connection
-connection = do
-  either (error . show) pure =<< HC.acquire connectionSettings
+-- | Prepare the database for use by Guide.
+--
+-- Determines which migrations have to be run, and runs them. Errors out if
+-- any migrations fail.
+--
+-- Does not create the @guide@ database if it does not exist yet. Use
+-- @CREATE DATABASE@ when running for the first time.
```
willbasky: > @CREATE DATABASE@

Is it touched upon the future?

neongreen: I'm not sure if we should create the database if it doesn't exist yet, so at the moment this is something you should do manually either by running `CREATE DATABASE guide;` or running Postgres with `POSTGRES_DB=guide`.

If creating the database automatically turns out to be the best practice, we can do it too, yes.

neongreen: CREATE DATABASE is an SQL command, not a function.

willbasky (commented):
OK

neongreen (approved):

PR #330, opened 2019-07-11
Use travis_retry with stack setup

neongreen:


.travis.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       before_script:
         - sudo apt-get install -y libgmp-dev
         - curl -sSL https://get.haskellstack.org/ | sh
-        - stack setup
+        - travis_retry stack setup
```
willbasky, line 49: From instruction: "If you are getting network timeouts when trying to download dependencies, either use the built in retry feature of your dependency manager or wrap your install commands in the travis_retry function."
My problem travis was with unfree directory to unpack (or install), it didn't concern network timeouts.

neongreen: Network timeouts are one usecase but I think it's not the only one.

The "unfree directory to unpack" thing is a Stack bug: https://github.com/commercialhaskell/stack/issues/4888. It happens for a different reason – if you look closely at the error message, it doesn't actually say that it failed _because_ the directory wasn't empty, it says that unpacking failed _and now_ the directory may not be empty.

willbasky: So, travis_retry is here for what reason? Just workaround?

willbasky (approved):
let's do it.

PR #332, opened 2019-07-12
Simplify client types

neongreen:
This PR gets rid of `?`, replacing it with explicit schema manipulation.

This also means that now FromJSON instances will accept missing fields as `Nothing` – aeson has built-in logic for treating `Maybe` fields like this, but it doesn't know anything about `Maybe ... ? ...`.

### DIFF
--- a/back/guide.cabal
+++ b/back/guide.cabal
@@ -96,10 +96,7 @@ library
     Guide.Routes
     Imports
     To
-  build-depends:       Spock
-                     , Spock-digestive
-                     , Spock-lucid
-                     , acid-state
+  build-depends:       acid-state
                      , aeson
                      , aeson-pretty
                      , async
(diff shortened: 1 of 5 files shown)


back/guide.cabal
```diff
@@ -96,10 +96,7 @@ library
     Guide.Routes
     Imports
     To
-  build-depends:       Spock
```
neongreen, line 99: Ignore these changes – it's churn caused by automatic sorting.

back/src/Guide/Api/Types.hs
```diff
@@ -800,13 +944,5 @@ instance ToSchema (Uid Trait) where
 
 instance ToSchema CategoryStatus
 
-instance ToSchema ItemKind where
```
neongreen, line 803: This instance wasn't needed anymore.

willbasky:
Why did you exclude tags?

neongreen:
What do you mean by "tags"?

If you mean "strings after `?`", then they are still available in Swagger, they just got moved into `ToSchema` instances.

willbasky (approved):

PR #333, opened 2019-07-12
Allow setting more properties (Hackage, etc) when creating an item

### LINKED ISSUE #325: Allow setting Hackage name right in the item creation dialog

neongreen:
fixes #325

### DIFF
--- a/back/src/Guide/Api/Types.hs
+++ b/back/src/Guide/Api/Types.hs
@@ -38,6 +38,7 @@ module Guide.Api.Types
   , CTrait(..), toCTrait
 
   -- * Request and response types
+  , CCreateItem(..)
   , CCreateTrait(..)
   , CCategoryInfoEdit(..)
   , CItemInfoEdit(..)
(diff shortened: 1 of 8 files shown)


back/src/Guide/Api/Types.hs
```diff
@@ -245,7 +246,7 @@ data TraitSite route = TraitSite
   ,  _createTrait :: route :-
       Summary "Create a new trait in the given item"
       :> Description "Returns the ID of the created trait."
-      :> ErrorResponse 400 "'content' not provided"
+      :> ErrorResponse 400 "'content' can not be empty"
```
neongreen, line 249: Would be nice to have some kind of `NotNullText` so that those would be handled automatically.

willbasky: let's open issue.

willbasky (approved):

back/src/Guide/Api/Types.hs
neongreen: I created an entry in https://www.notion.so/aelve/Backend-technical-debt-45a41ab060b7400bb7d0e6547fa34785

neongreen:
@avele rebase and merge when ready

PR #339, opened 2019-07-15
Add get functions for categories, traits and items

willbasky:


back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,87 @@
...
+      decoder = HD.rowMaybe (Item
+        <$> (Uid <$> (HD.column . HD.nonNullable) HD.text)                             -- uid
+        <*> ((HD.column . HD.nonNullable) HD.text)                                     -- name
+        <*> (localTimeToUTC utc <$> (HD.column . HD.nonNullable) HD.timestamp)         -- localtime
+        <*> ((HD.column . HD.nullable) HD.text)                                        -- group_
+        <*> ((HD.column . HD.nullable) HD.text)                                        -- link
+        <*> ((HD.column . HD.nullable) HD.text)                                        -- hackage
+        <*> (toMarkdownBlock <$> (HD.column . HD.nonNullable) HD.text)                 -- summary
+        <*> (toMarkdownBlock <$> (HD.column . HD.nonNullable) HD.text)                 -- ecosystem
+        <*> (toMarkdownTree "whatisidprefix" <$> (HD.column . HD.nonNullable) HD.text) -- notes. What a prefix?
```
willbasky: @neongreen What is it prefix about?

back/src/Guide/Types/Core.hs
```diff
@@ -178,15 +178,15 @@ data Item = Item {
...
+  _itemNotes       :: MarkdownTree,    -- ^ The notes section
   _itemPros        :: [Trait],         -- ^ Pros (positive traits)
   _itemProsDeleted :: [Trait],         -- ^ Deleted pros go here (so that
                                        --   it'd be easy to restore them)
   _itemCons        :: [Trait],         -- ^ Cons (negative traits)
-  _itemConsDeleted :: [Trait],         -- ^ Deleted cons go here
-  _itemEcosystem   :: MarkdownBlock,   -- ^ The ecosystem section
-  _itemNotes       :: MarkdownTree,    -- ^ The notes section
-  _itemLink        :: Maybe Url        -- ^ Link to homepage or something
+  _itemConsDeleted :: [Trait]          -- ^ Deleted cons go here
```
willbasky: This permutations for consistenting with Postgres schema.

back/src/Guide/Db/Get.hs
neongreen: https://github.com/aelve/guide/blob/develop/back/src/Guide/State.hs#L388-L389

The prefix specifies how `id`s for headers should look in rendered HTML. Let's say two items' notes both look like this:

```markdown
# Problems
...

# Solutions
...
```

If we generate `<h1 id="problems">` for both of those, linking to a particular part of a particular note with `guide.aelve.com/haskell/someid#problems` becomes impossible. The solution is to have a different prefix for all headers in notes. Then links will look like this: `guide.aelve.com/haskell/someid#item-notes-someid-problems`.

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,87 @@
...
+getItemByItemIdMaybe :: Uid Item -> Session (Maybe Item)
+getItemByItemIdMaybe uid@(Uid uidItem) = do
+  pro <- getTraitByItemIdMaybe False "pro" uid
+  proDeleted <- getTraitByItemIdMaybe True "pro" uid
+  con <- getTraitByItemIdMaybe False "con" uid
+  conDeleted <- getTraitByItemIdMaybe True "con" uid
+  let sql = toByteString [text|
+        SELECT * FROM items WHERE uid = '${uidItem}';|]
+      encoder = HE.noParams
+      decoder = HD.rowMaybe (Item
```
neongreen: Please use `-XApplicativeDo` so that you wouldn't accidentally reorder the columns.

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,87 @@
...
+
+-- | Get item by id, either it deleted or not.
+getItemByItemIdMaybe :: Uid Item -> Session (Maybe Item)
+getItemByItemIdMaybe uid@(Uid uidItem) = do
+  pro <- getTraitByItemIdMaybe False "pro" uid
+  proDeleted <- getTraitByItemIdMaybe True "pro" uid
+  con <- getTraitByItemIdMaybe False "con" uid
+  conDeleted <- getTraitByItemIdMaybe True "con" uid
+  let sql = toByteString [text|
+        SELECT * FROM items WHERE uid = '${uidItem}';|]
```
neongreen: Don't ever use interpolation to construct SQL statements! It's fine in this case because `uidItem` can not contain single quotes, but in general this is exactly why SQL injections happen.

In fact, to prevent this from happening let's switch to https://hackage.haskell.org/package/raw-strings-qq-1.1/docs/Text-RawString-QQ.html for SQL statements. As a bonus, you won't need `toByteString` anymore.

To pass parameters to SQL queries, use `$1`, `$2`, etc. See http://hackage.haskell.org/package/hasql-1.4/docs/Hasql-Statement.html.

neongreen: > `SELECT *`

Please list the columns explicitly, in my experience the improved greppability is worth it.

back/src/Guide/Types/Core.hs
neongreen: I like changing the order for consistency, but changing the order would require an acid-state migration, which is annoying. Can you leave the order as-is and change the schema order instead?

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,87 @@
...
+        <$> (Uid <$> (HD.column . HD.nonNullable) HD.text)
+        <*> (toMarkdownInline <$> (HD.column . HD.nonNullable) HD.text))
+  HS.statement () (Statement sql encoder decoder False)
+
+-- | Get item by id, either it deleted or not.
+getItemByItemIdMaybe :: Uid Item -> Session (Maybe Item)
+getItemByItemIdMaybe uid@(Uid uidItem) = do
+  pro <- getTraitByItemIdMaybe False "pro" uid
+  proDeleted <- getTraitByItemIdMaybe True "pro" uid
+  con <- getTraitByItemIdMaybe False "con" uid
```
neongreen: Don't use naked booleans, it's hard to guess what this code does. Use [named](http://hackage.haskell.org/package/named).

neongreen: `"pro"` and `"con"` should be replaced by an ADT in code.

back/src/Guide/Types/Core.hs
willbasky: of course

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,209 @@
...
+        _itemName <- textRowNonNull
+        _itemCreated <- localTimeToUTC utc <$> (HD.column . HD.nonNullable) HD.timestamp
+        _itemGroup_ <- textRowNull
+        _itemLink <- textRowNull
+        _itemHackage <- textRowNull
+        _itemSummary <- toMarkdownBlock <$> textRowNonNull
+        _itemEcosystem <- toMarkdownBlock <$> textRowNonNull
+        _itemNotes <- toMarkdownTree pref <$> textRowNonNull
+
+        -- How-to get procon with item uid fetched just now?
```
willbasky: :thinking:

neongreen:
I pushed a small commit

back/src/Guide/Db/Connection.hs
```diff
@@ -0,0 +1,48 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module Guide.Db.Connection
```
neongreen: All modules should have haddocks.

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,305 @@
+{-# LANGUAGE ApplicativeDo     #-}
+{-# LANGUAGE DataKinds         #-}
+{-# LANGUAGE OverloadedLabels  #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes       #-}
+{-# LANGUAGE TypeOperators     #-}
+{-# LANGUAGE ViewPatterns      #-}
+
+module Guide.Db.Get where
```
neongreen: This module doesn't have an export list.

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,305 @@
...
+
+-- | Get traits list by item id, deleted and trait_type filters.
+getTraitsByItem :: Uid Item -> "deleted" :! Bool -> TraitType -> Session [Trait]
+getTraitsByItem Uid{..} (arg #deleted -> deleted) traitType = do
+  let sql = [r|
+        SELECT uid, content FROM traits
+          WHERE item_uid = $1 AND
+                deleted = $2 AND
+                type_ = ($3 :: trait_type)
+        ;|]
```
neongreen: Are semicolons necessary? It would be nicer not to have them.

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,305 @@
...
+  let sql = [r|SELECT uid, content FROM traits WHERE uid = $1;|]
+      encoder = textEncoderNonNull
+      decoder = HD.rowMaybe traitRow
+  HS.statement uidToText (Statement sql encoder decoder False)
+
+-- | Get traits list by item id, deleted and trait_type filters.
+getTraitsByItem :: Uid Item -> "deleted" :! Bool -> TraitType -> Session [Trait]
+getTraitsByItem Uid{..} (arg #deleted -> deleted) traitType = do
+  let sql = [r|
+        SELECT uid, content FROM traits
```
neongreen: SQL should be formatted like SQL, not like Haskell.

My understanding of SQL style conventions is that this query should look like this:

```sql
SELECT uid, content
FROM traits
WHERE item_uid = $1
  AND deleted = $2
  AND type_ = ($3 :: trait_type)
```

neongreen: By the way, the `:: trait_type` syntax is unfamiliar – can you send a link to the docs?

back/src/Guide/Db/Get.hs
neongreen: Btw, let's rename the hierarchy to "Guide.Database", it's more reader-friendly.

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,305 @@
...
+  print cat
+  catId <- run' (getCategoryIdByItem "items1234567") conn
+  print catId
+  catIds <- run' getCategoryIds conn
+  print catIds
+  cats <- run' getCategories conn
+  print cats
+
+--
+-- | Methods to get information from database
```
neongreen: We use a different comment style in the codebase.

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,305 @@
...
+--
+
+--
+-- | Traits
+--
+
+-- | Get trait by id, either it deleted or not.
+getTraitMaybe :: Uid Trait -> Session (Maybe Trait)
+getTraitMaybe Uid{..} = do
+  let sql = [r|SELECT uid, content FROM traits WHERE uid = $1;|]
```
neongreen: Let's write all queries multi-line.

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,305 @@
...
+      encoder = textEncoderNonNull
+      decoder = HD.rowMaybe traitRow
+  HS.statement uidToText (Statement sql encoder decoder False)
+
+-- | Get traits list by item id, deleted and trait_type filters.
+getTraitsByItem :: Uid Item -> "deleted" :! Bool -> TraitType -> Session [Trait]
+getTraitsByItem Uid{..} (arg #deleted -> deleted) traitType = do
```
neongreen: It's hard to guess where `uidToText` comes from. You can write `(Uid uidText)`. Better still, let's have a `uidEncoder` and avoid unwrapping UIDs.

neongreen: `Pro` and `Con` are unclear at call site. Let's use `named` for `TraitType` as well so that the callsite would look like:

```haskell
(getTraitsByItem "items1234567" (#deleted False) (#traitType Pro))
```

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,305 @@
...
+
+--
+-- | Encoders
+--
+
+-- | Encoder parameter for text
+textEncoderNonNull :: HE.Params Text
+textEncoderNonNull = HE.param (HE.nonNullable HE.text)
+
+-- | Encoder parameter for text
```
neongreen: Copypaste typo

back/src/Guide/Db/Schema.hs
```diff
@@ -136,17 +101,17 @@ v0 = do
...
 
 -- | Create table @traits@, corresponding to 'Guide.Types.Core.Trait'.
 v0_createTableTraits :: Session ()
-v0_createTableTraits = HS.sql $ toByteString [text|
+v0_createTableTraits = HS.sql [r|
   CREATE TABLE traits (
     uid text PRIMARY KEY,           -- Unique trait ID
     content text NOT NULL,          -- Trait content as Markdown
-      deleted boolean               -- Whether the trait is deleted
+    deleted boolean               -- Whether the trait is deleted
```
neongreen:
```suggestion
    deleted boolean                 -- Whether the trait is deleted
```

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,305 @@
...
+
+-- | Trait row
+traitRow :: HD.Row Trait
+traitRow = do
+  _traitUid <- Uid <$> textDecoderNonNull
+  _traitContent <- toMarkdownInline <$> textDecoderNonNull
+  pure $ Trait{..}
+
+-- | Item row
+itemRow :: Text -> [Trait] -> [Trait] -> [Trait] -> [Trait] -> HD.Row Item
```
neongreen: The `itemRow` function depends on the order of parameters in `SELECT`, and therefore it should not be a standalone function. You can move it into `getItemMaybe`.

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,305 @@
...
+-- | Encoder parameter for bool
+boolEncoderNonNull :: HE.Params Bool
+boolEncoderNonNull = HE.param (HE.nonNullable HE.bool)
+
+--
+-- | Decoders
+--
+
+-- | Decoder for nonNull text
+textDecoderNonNull :: HD.Row Text
```
neongreen: I'd like to move all those auxiliary encoders and decoders to a separate module. Say, `Guide.Database.Convert`.

Given that most of our columns are non-nullable, I'd propose the following naming scheme:

```haskell
boolEncoder = HE.nonNullable HE.bool
boolEncoderNullable = HE.nullable HE.bool
boolParam = HE.param boolEncoder
boolColumn = HE.column boolEncoder
```

`textDecoderNull` should go away – it's not a `Decoder`, it's a `Row`.

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,305 @@
...
+        SELECT uid, name, created, group_, link, hackage, summary, ecosystem, notes
+          FROM items
+          WHERE uid = $1;
+        |]
+      encoder = textEncoderNonNull
+      decoder = HD.rowMaybe $
+        itemRow pref _itemPros _itemProsDeleted _itemCons _itemConsDeleted
+  HS.statement uidToText (Statement sql encoder decoder False)
+
+-- | Get item by id, either it deleted or not.
```
neongreen: This function should just use `getItemMaybe`.

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,305 @@
...
+  _traitUid <- Uid <$> textDecoderNonNull
+  _traitContent <- toMarkdownInline <$> textDecoderNonNull
+  pure $ Trait{..}
+
+-- | Item row
+itemRow :: Text -> [Trait] -> [Trait] -> [Trait] -> [Trait] -> HD.Row Item
+itemRow pref _itemPros _itemProsDeleted _itemCons _itemConsDeleted = do
+  _itemUid <- Uid <$> textDecoderNonNull
+  _itemName <- textDecoderNonNull
+  _itemCreated <- localTimeToUTC utc <$> (HD.column . HD.nonNullable) HD.timestamp
```
neongreen: It turns out that `timestamptz` doesn't actually store the time zone – it's more like UTCTime. Let's use `timestamptz` here and in the database schema.

https://dba.stackexchange.com/questions/59006/what-is-a-valid-use-case-for-using-timestamp-without-time-zone

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,305 @@
...
+
+-- | Get all category's items
+categoryItems :: Uid Category -> Session ([Item], [Item])
+categoryItems catId = do
+  _categoryItems <- getItemsByCategory catId (#deleted False)
+  _categoryItemsDeleted <- getItemsByCategory catId (#deleted True)
+  pure (_categoryItems, _categoryItemsDeleted)
+
+-- | Just ADT for traitType
+data TraitType = Pro | Con
```
neongreen: This should live in `Guide.Types.Core`.

back/src/Guide/Db/Get.hs
```diff
@@ -0,0 +1,305 @@
...
+categoryRow :: [Item] -> [Item] -> HD.Row Category
+categoryRow _categoryItems _categoryItemsDeleted = do
+  _categoryUid <- Uid <$> textDecoderNonNull
+  _categoryTitle <- textDecoderNonNull
+  _categoryCreated <- localTimeToUTC utc <$> (HD.column . HD.nonNullable) HD.timestamp
+  _categoryGroup_ <- textDecoderNonNull
+  _categoryStatus <- toStatus <$> textDecoderNonNull
+  _categoryNotes <- toMarkdownBlock <$> textDecoderNonNull
+  selectionsText <- textArrayDecoder
+  let _categoryEnabledSections = Set.fromList $ map toSection selectionsText
```
neongreen: I don't like `toSection` and `toStatus` – they kinda depend on our internal type names. I'd rather have conversions inside `Guide.Database`. While I'm at it, we can have just "stub", "wip", "finished", there's no need for CategoryStatusBlahBlahStub.

back/src/Guide/Db/Get.hs
willbasky: > can you send a link to the docs?

it was advice from `Hasql` author in PM.

back/src/Guide/Db/Get.hs
willbasky: What to return if case Nothing?
I fail it

.gitignore
```diff
@@ -22,6 +22,8 @@ cabal.config
 *.swp
 *.swo
 *#
+*yaml.lock
```
neongreen: Why not just *.lock?

```suggestion
*.lock
```

back/src/Guide/Database/Connection.hs
```diff
@@ -0,0 +1,52 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+
+--
+-- Connection methods for postgres database
```
neongreen:
```suggestion
-- | Connection methods for postgres database
```

back/src/Guide/Database/Convert.hs
```diff
@@ -0,0 +1,173 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+
+module Guide.Database.Convert
+       (
+       -- Encoders
```
neongreen: Export list sections should look like `-- * Encoders`, `-- ** Encoders`, etc. Otherwise they won't show up in Haddock.

back/src/Guide/Database/Convert.hs
```diff
@@ -0,0 +1,173 @@
...
+         boolParam
+       , boolParamNullable
+       , textParam
+       , textParamNullable
+       , traitTypeParam
+       , uidParam
+       -- Decoders
```
neongreen: I think it's better to group functions by type (bool/text/etc), not by decoder/encoder, because then it will be easier to notice if an encoder and a decoder don't match.

back/src/Guide/Database/Convert.hs
```diff
@@ -0,0 +1,173 @@
...
+import qualified Hasql.Decoders as HD
+import qualified Hasql.Encoders as HE
+
+import Guide.Markdown (toMarkdownBlock, toMarkdownInline, toMarkdownTree)
+import Guide.Types.Core (Category (..), CategoryStatus(..), Item (..), ItemSection(..), Trait (..),
+                         TraitType (..))
+import Guide.Utils (Uid (..))
+
+
+-- Convert functions for database
```
neongreen: This should be a module haddock (`-- |` before `module`), not a stray comment after imports.

back/src/Guide/Database/Convert.hs
```diff
@@ -0,0 +1,173 @@
...
+
+-- | Decode Trait
+traitRow :: HD.Row Trait
+traitRow = do
+  _traitUid <- uidColumn
+  _traitContent <- toMarkdownInline <$> textColumn
+  pure $ Trait{..}
+
+-- | Decode Item
+itemRow :: Text -> [Trait] -> [Trait] -> [Trait] -> [Trait] -> HD.Row Item
```
neongreen: As I said, these functions depend on the order of columns in `SELECT` and therefore they should be in the same place as the `SELECT`.

back/src/Guide/Database/Convert.hs
```diff
@@ -0,0 +1,173 @@
...
+  let _categoryEnabledSections = Set.fromList $ map toSection selectionsText
+  let _categoryGroups = Map.empty
+  pure $ Category{..}
+
+----------------------------------------------------------------------------
+-- Other
+----------------------------------------------------------------------------
+
+-- | Read text to CategoryStatus
+toStatus :: Text -> CategoryStatus
```
neongreen: You can do the same for `TraitType`. In general it's best not to rely on `Show` for anything except logging.

back/src/Guide/Database/Get.hs
```diff
@@ -0,0 +1,219 @@
...
+  cat <- run' (getCategory "categories11") conn
+  print cat
+  catId <- run' (getCategoryIdByItem "items1234567") conn
+  print catId
+  catIds <- run' getCategoryIds conn
+  print catIds
+  cats <- run' getCategories conn
+  print cats
+
+----------------------------------------------------------------------------
```
neongreen, line 80: Empty section.

back/src/Guide/Database/Convert.hs
```diff
@@ -0,0 +1,173 @@
...
+toStatus = \case
+  "stub" -> CategoryStub
+  "wip" -> CategoryWIP
+  "finished" -> CategoryFinished
+  _ -> error "Unknown CategoryStatus"
```
neongreen: Those should not `error` out – instead we should use whatever mechanism Hasql itself uses for throwing errors when it can't parse the row.

back/src/Guide/Database/Schema.hs
```diff
@@ -2,21 +2,23 @@
 {-# LANGUAGE QuasiQuotes       #-}
 
 
-module Guide.Db.Schema
+--
+-- | Schemas to create table for guide database
```
neongreen, line 5: This PR introduces three different styles for module comments. This is inconsistent.

The comment style we should use is:

```haskell
-- | Foo bar baz.
-- Blah blah blah.
module Foo where
```

back/src/Guide/Database/Schema.hs
```diff
@@ -211,7 +179,7 @@ v0_createTableUsers = HS.sql $ toByteString [text|
 -- | Create table @pending_edits@, storing users' edits and metadata about
 -- them (who made the edit, when, etc).
 v0_createTablePendingEdits :: Session ()
-v0_createTablePendingEdits = HS.sql $ toByteString [text|
+v0_createTablePendingEdits = HS.sql [r|
```
neongreen, line 181: This table should also use `timestamptz` instead of `timestamp`. Please be more careful when grepping.

back/src/Guide/Api/Types.hs
```diff
@@ -308,17 +308,27 @@ type Api = RequestDetails :> ToServant Site AsApi
...
-data CTraitType = Pro | Con
+data CTraitType = CPro | CCon
     deriving (Show, Generic)
 
 instance ToSchema CTraitType where
     declareNamedSchema = genericDeclareNamedSchema schemaOptions
+      { constructorTagModifier = \case
+          "CPro" -> "Pro"
+          "CCon" -> "Con"
+          other -> error ("TraitType schema: unknown tag " <> show other)
```
neongreen:
```suggestion
          other -> error ("СTraitType schema: unknown tag " <> show other)
```

back/src/Guide/Api/Types.hs
```diff
@@ -308,17 +308,27 @@ type Api = RequestDetails :> ToServant Site AsApi
...
 
 instance A.ToJSON CTraitType where
-  toJSON = A.genericToJSON jsonOptions
+  toJSON = \case
+    CPro -> "Pro"
+    CCon -> "Con"
 
 instance A.FromJSON CTraitType where
-  parseJSON = A.genericParseJSON jsonOptions
+  parseJSON = \case
```
neongreen: Use `withText`.

back/src/Guide/Database/Connection.hs
```diff
@@ -0,0 +1,61 @@
...
+----------------------------------------------------------------------------
+-- Utilities
+----------------------------------------------------------------------------
+
+-- | Like 'HS.run', but errors out in case of failure. For DatabaseMonad
+runDM :: DatabaseMonad a -> Connection -> IO a
+runDM s c = eitherRun =<< eitherRun =<< HS.run (runExceptT s) c
+
+-- | Like 'HS.run', but errors out in case of failure. For Session
+runS :: Session a -> Connection -> IO a
```
neongreen: Non-abbreviated names are better. `runSession`

back/src/Guide/Database/Types.hs
```diff
@@ -0,0 +1,27 @@
...
+data DatabaseError
+  = ItemNotFound (Uid Item)
+  | CategoryNotFound (Uid Category)
+  deriving Show
+
+-- | ExceptT wrapper on Session with custom errors
+type DatabaseMonad = ExceptT DatabaseError Session
+
+-- | Convert Session to DatabaseMonad
+toDatabaseMonad :: Session a -> ExceptT DatabaseError Session a
```
neongreen: This is `lift`, you can remove this function.

back/src/Guide/Database/Types.hs
```diff
@@ -0,0 +1,27 @@
...
+import Hasql.Session (Session)
+
+-- | Custom datatype errors for database
+data DatabaseError
+  = ItemNotFound (Uid Item)
+  | CategoryNotFound (Uid Category)
+  deriving Show
+
+-- | ExceptT wrapper on Session with custom errors
+type DatabaseMonad = ExceptT DatabaseError Session
```
neongreen: I would actually suggest removing this type synonym and just write `ExceptT DatabaseError Session`, then it's more immediately obvious what's going on.

Full disclosure: another reason is that the name isn't great and I can't think of a good one.

back/src/Guide/Database/Get.hs
```diff
@@ -0,0 +1,235 @@
...
+        _itemNotes <- toMarkdownTree pref <$> textColumn
+        pure $ Item{..}
+  toDatabaseMonad $ HS.statement itemId (Statement sql encoder decoder False)
+
+-- | Get item by id, either it deleted or not.
+getItem :: Uid Item -> DatabaseMonad Item
+getItem itemId = do
+  mItem <- getItemMaybe itemId
+  case mItem of
+        Nothing   -> ExceptT $ pure $ Left $ ItemNotFound itemId
```
neongreen: `Nothing -> throwError (ItemNotFound itemId)`

back/src/Guide/Database/Get.hs
```diff
@@ -0,0 +1,235 @@
...
+        Just item -> ExceptT $ pure $ Right item
```
neongreen: `Just item -> pure item`

back/src/Guide/Database/Get.hs
```diff
@@ -0,0 +1,235 @@
...
+        let _categoryGroups = Map.empty
+        pure $ Category{..}
+  toDatabaseMonad $ HS.statement catId (Statement sql encoder decoder False)
+
+-- | Get category by uid.
+getCategory :: Uid Category -> DatabaseMonad Category
+getCategory catId = do
+  mCat <- getCategoryMaybe catId
+  case mCat of
+    Nothing  -> fail "getCategory returns nothing"
```
neongreen: This should use `CategoryNotFound`.

back/src/Guide/Database/Types.hs
willbasky: SessionExceptT
or
ExceptTSession

back/src/Guide/Api/Types.hs
```diff
@@ -325,9 +325,10 @@ instance A.ToJSON CTraitType where
     CCon -> "Con"
 
 instance A.FromJSON CTraitType where
-  parseJSON traitType =
-        A.withText "Pro" (const (pure CPro)) traitType
-    <|> A.withText "down" (const (pure CCon)) traitType
+  parseJSON = A.withText "CTraitType" $ \case
+    "Pro" -> pure CPro
+    "Con" -> pure CCon
+    other -> fail ("unknown trait type " <> show other)
```
willbasky, line 331: Why was it returned?

PR #355, opened 2019-07-29
Add insert/update transactions

willbasky:


back/src/Guide/Database/Add.hs
```diff
@@ -0,0 +1,158 @@
...
+import Guide.Database.Types
+import Guide.Types.Core (Category (..), CategoryStatus (..), Item (..), Trait (..), TraitType (..))
+import Guide.Utils (Uid (..))
+
+
+-- | Insert category to database.
+addCategory
+  :: Uid Category       -- ^ New category's id
+  -> "title" :! Text    -- ^ Title
+  -> "group_" :! Text   -- ^ Group
```
neongreen: It can be named just `group`, that's fine. `#group` and `group` don't clash.

back/src/Guide/Database/Add.hs
```diff
@@ -0,0 +1,158 @@
...
+  -> UTCTime            -- ^ Creation time
+  -> ExceptT DatabaseError Transaction ()
+addCategory catId (arg #title -> title) (arg #group_ -> group_) created = do
+  let sql = [r|
+        INSERT INTO categories (uid, title, created, group_, status_, notes, enabled_sections, items_order)
```
neongreen: Actually, do you remember why we called it `status_`? It's not a reserved keyword as far as I know: https://www.postgresql.org/docs/11/sql-keywords-appendix.html.

back/src/Guide/Database/Convert.hs
```diff
@@ -18,16 +18,24 @@ module Guide.Database.Convert
        , uidParam
        , uidColumn
 
+       -- * '[Uid]
```
neongreen: The `'Foo'` syntax only works for linking identifiers. If you write `'[Foo]'`, Haddock won't do anything with it. You can write `['Foo']` instead.

back/src/Guide/Database/Convert.hs
```diff
@@ -92,6 +100,10 @@ uidColumn = Uid <$> textColumn
 -- UTCTime
 ----------------------------------------------------------------------------
 
+-- | Encode a 'UTCTime'.
```
neongreen:
```suggestion
-- | Pass a 'UTCTime' to a query.
```

back/src/Guide/Database/Convert.hs
```diff
@@ -147,3 +197,22 @@ itemSectionSetColumn =
...
+
+-- | Pass a @[Uid]@ to a query.
+uidsParam :: HE.Params [Uid a]
+uidsParam = HE.param
+  $ HE.nonNullable
+  $ HE.foldableArray
+  $ HE.nonNullable
+  $ contramap uidToText HE.text
+
+uidsColumn :: HD.Row [Uid a]
```
neongreen, line 249: Haddock.

back/src/Guide/Database/Convert.hs
```diff
@@ -147,3 +197,22 @@ itemSectionSetColumn =
...
+----------------------------------------------------------------------------
+-- [Uid]
+----------------------------------------------------------------------------
+
+-- | Pass a @[Uid]@ to a query.
+uidsParam :: HE.Params [Uid a]
+uidsParam = HE.param
+  $ HE.nonNullable
+  $ HE.foldableArray
+  $ HE.nonNullable
```
neongreen: You can use `uidParam` here.

back/src/Guide/Database/Convert.hs
```diff
@@ -147,3 +197,22 @@ itemSectionSetColumn =
...
+  $ contramap uidToText HE.text
+
+uidsColumn :: HD.Row [Uid a]
+uidsColumn = HD.column
+  $ HD.nonNullable
+  $ HD.listArray
+  $ HD.nonNullable
```
neongreen: `uidColumn`

back/src/Guide/Database/Get.hs
```diff
@@ -122,6 +97,19 @@ getTraitsByItem itemId (arg #deleted -> deleted) (arg #traitType -> traitType) =
         pure $ Trait{..}
   lift $ HT.statement (itemId,deleted,traitType) (Statement sql encoder decoder False)
 
+-- | Get a 'TraitType' by traitId.
+getTraitTypeByTraitId :: Uid Trait -> ExceptT DatabaseError Transaction TraitType
```
neongreen: What is this one for?

back/src/Guide/Database/Get.hs
```diff
@@ -105,9 +80,9 @@ getTraitMaybe traitId = do
 getTraitsByItem
   :: Uid Item
   -> "deleted" :! Bool
-  -> "traitType" :! TraitType
```
neongreen, line 108: Why?

back/src/Guide/Database/Schema.hs
```diff
@@ -140,7 +140,11 @@ v0_createTableItems = HS.sql [r|
       NOT NULL,
     category_uid text               -- Category that the item belongs to
       REFERENCES categories (uid)
-      ON DELETE CASCADE
+      ON DELETE CASCADE,
+    pros_order text[]               -- Order of pro traits
```
neongreen: You should also write that the list contains uids.

back/src/Guide/Database/Set.hs
```diff
@@ -0,0 +1,584 @@
...
+  let sql = [r|
+        UPDATE categories
+        SET notes = $2
+        WHERE uid = $1
+        |]
+      encoder = contrazip2 uidParam textParam
+      decoder = HD.noResult
+  lift $ HT.statement (catId, notes) (Statement sql encoder decoder False)
+
+-- | Set new category notes.
```
neongreen: Typo.

back/src/Guide/Database/Set.hs
```diff
@@ -0,0 +1,584 @@
...
+        WHERE uid = $1
+        |]
+      encoder = contrazip2 uidParam categoryStatusParam
+      decoder = HD.noResult
+  lift $ HT.statement (catId, status) (Statement sql encoder decoder False)
+
+-- | Set category selections.
+setCategoryEnabledSections
+  :: Uid Category
+  -> Set ItemSection     -- ^ Sections to enable
```
neongreen: This should definitely use `Named`.

back/src/Guide/Database/Set.hs
```diff
@@ -0,0 +1,584 @@
...
+      encoder = contrazip2 uidParam itemSectionSetParam
+      decoder = HD.noResult
+  lift $ HT.statement (catId, newSections) (Statement sql encoder decoder False)
+
+-- | Add itemId to category items_order.
+addItemIdToCategory :: Uid Category -> Uid Item -> ExceptT DatabaseError Transaction ()
+addItemIdToCategory catId itemId = do
+    itemsOrder <- getCategoryItemsOrder catId
+    let addItemId =
+          if notElem itemId itemsOrder then itemId : itemsOrder else itemsOrder
```
neongreen: If the item was already in `items_order`, it's always a bug. It should be reported and not thrown away.

back/src/Guide/Database/Set.hs
```diff
@@ -0,0 +1,584 @@
...
+        WHERE uid = $1
+        |]
+      encoder = contrazip2 uidParam textParamNullable
+      decoder = HD.noResult
+  lift $ HT.statement (itemId, link) (Statement sql encoder decoder False)
+
+-- | Set item hackage.
+--
+-- "hackage" :? Text == Maybe Text with label.
+setItemHackage :: Uid Item -> "hackage" :? Text -> ExceptT DatabaseError Transaction ()
```
neongreen: No, use `"hackage" :! Maybe Text`. The `:?` operator is used for optional arguments, and the Hackage argument is a mandatory parameter, just one that we can either set to `Just something` or to `Nothing`.

back/src/Guide/Database/Set.hs
```diff
@@ -0,0 +1,584 @@
...
+      decoder = HD.noResult
+  lift $ HT.statement (itemId, ecosystem) (Statement sql encoder decoder False)
+
+-- | Set item to be deleted.
+--
+-- When 'deleted' is True, item will marked as deleted (_categoryItemsDeleted)
+-- otherwise item will marked as notDeleted (_categoryItems)
+setItemDeleted :: Uid Item -> "deleted" :! Bool -> ExceptT DatabaseError Transaction ()
+setItemDeleted itemId (arg #deleted -> deleted) = do
+  mCategoryId <- getCategoryIdByItemMaybe itemId
```
neongreen: Let's have a `getCategoryIdByItem` that will throw `ItemNotFound` when necessary. This way you won't have to handle the `Maybe` and thrown errors will be more precise. `ItemOrCategoryNotFound` is a bad error, it doesn't help with debugging much.

back/src/Guide/Database/Set.hs
```diff
@@ -0,0 +1,584 @@
...
+            WHERE uid = $1
+            |]
+          encoder = contrazip2 uidParam boolParam
+          decoder = HD.noResult
+      lift $ HT.statement (itemId, deleted) (Statement sql encoder decoder False)
+      if deleted then deleteItemIdFromCategory catId itemId
+      else addItemIdToCategory catId itemId
+
+-- | Move item up or down.
+moveItem :: Uid Item -> "direction" :! Bool -> ExceptT DatabaseError Transaction ()
```
neongreen: Let's have a separate type for direction.

back/src/Guide/Database/Set.hs
```diff
@@ -0,0 +1,584 @@
...
+    Nothing -> throwError TraitOrItemNotFound
+    Just itemId -> do
+      let sql = [r|
+            DELETE FROM traits
+            WHERE uid = $1
+            |]
+          encoder = uidParam
+          decoder = HD.noResult
+      lift $ HT.statement traitId (Statement sql encoder decoder False)
+      traitType <- getTraitTypeByTraitId traitId
```
neongreen: This won't work – you are querying the trait after deleting it.

back/src/Guide/Database/Types.hs
```diff
@@ -14,4 +14,6 @@ import Guide.Utils (Uid (..))
 data DatabaseError
   = ItemNotFound (Uid Item)
   | CategoryNotFound (Uid Category)
+  | ItemOrCategoryNotFound
```
neongreen: These are non-descriptive and should be removed if possible.

back/src/Guide/Database/Set.hs
```diff
@@ -0,0 +1,584 @@
...
+  let sql = [r|
+        UPDATE items
+        SET hackage = $2
+        WHERE uid = $1
+        |]
+      encoder = contrazip2 uidParam textParamNullable
+      decoder = HD.noResult
+  lift $ HT.statement (itemId, hackage) (Statement sql encoder decoder False)
+
+-- | Set item summary.
```
neongreen: - [ ] TODO: add a note that this function should not be used without doing a `get` first in the same transaction. I will add the note later.

back/src/Guide/Database/Get.hs
willbasky: TraitType is the same as "traitType". Name messege us what is it.

back/src/Guide/Database/Get.hs
willbasky: it used in `setTraitDeleted` at Set module.

back/src/Guide/Database/Add.hs
willbasky: Something highlighted a clash with `status`, i guess.

back/src/Guide/Database/Add.hs
```diff
@@ -0,0 +1,258 @@
...
+-- | Insert trait to database.
+addTrait
+  :: Uid Item             -- ^ Item id
+  -> Uid Trait            -- ^ New trait's id
+  -> TraitType            -- ^ Pro or Con
+  -> "content" :! Text    -- ^ Trait content
+  -> ExceptT DatabaseError Transaction ()
+addTrait itemId traitId traitType (arg #content -> content) = do
+  let sql = [r|
+        INSERT INTO traits (uid, content, type_, item_uid)
```
neongreen: Won't work. The row contains five parameters so you should add `deleted` too.

Come to think of it, we will never use the default for `deleted` so we might just as well remove it from the schema.

back/src/Guide/Database/Get.hs
neongreen: > Name messege us what is it.

Sure, but now you don't get documentation at the call site – you need to know what `Pro` and `Con` are, and they are kinda vague and non-descriptive names.

back/src/Guide/Database/Get.hs
neongreen: You can just fetch the whole trait in `setTraitDeleted`, then. It's better than to have a separate function for every column we might want to fetch.

back/src/Guide/Database/Get.hs
```diff
@@ -217,6 +247,18 @@ getCategoryMaybe catId = do
...
+-- | Get 'items_order' of category.
+getCategoryItemsOrder :: Uid Category -> ExceptT DatabaseError Transaction [Uid Item]
+getCategoryItemsOrder catId = do
+  let sql = [r|
+        SELECT items_order
+        FROM categories
+        WHERE uid = $1
+        |]
+      encoder = uidParam
+      decoder = HD.singleRow uidsColumn
```
neongreen: `singleRow` can fail. Use `rowMaybe` and throw `CategoryNotFound` when necessary.

back/src/Guide/Database/Get.hs
neongreen: Though then we should make sure that we fetch only the row and don't do any expensive conversion (like parsing Markdown). I'll look into that.

back/src/Guide/Database/Get.hs
neongreen: I will rename them.

back/src/Guide/Database/Get.hs
```diff
@@ -188,6 +178,46 @@ getItemsByCategory catId (arg #deleted -> deleted) = do
...
+        FROM traits
+        WHERE uid = $1
+        |]
+      encoder = uidParam
+      decoder = HD.rowMaybe $ uidColumn
+  lift $ HT.statement traitId (Statement sql encoder decoder False)
+
+-- | Get items by trait
+--
+-- Throw an error when 'Uid Trait' not found.
```
neongreen: Which error? It's better to mention that in the Haddock.

`-- Can throw 'TraitNotFound'.`

back/src/Guide/Database/Set.hs
```diff
@@ -0,0 +1,570 @@
...
+        UPDATE categories
+        SET title = $2
+        WHERE uid = $1
+        |]
+      encoder = contrazip2 uidParam textParam
+      decoder = HD.noResult
+  lift $ HT.statement (catId, title) (Statement sql encoder decoder False)
+
+-- | Set category group.
+setCategoryGroup :: Uid Category -> "group_" :! Text -> ExceptT DatabaseError Transaction ()
```
neongreen: You forgot to rename the `"group_"` here.

back/src/Guide/Database/Set.hs
```diff
@@ -0,0 +1,570 @@
...
+        |]
+      encoder = contrazip2 uidParam itemSectionSetParam
+      decoder = HD.noResult
+  lift $ HT.statement (catId, newSections) (Statement sql encoder decoder False)
+
+-- | Add itemId to category items_order.
+addItemIdToCategory :: Uid Category -> Uid Item -> ExceptT DatabaseError Transaction ()
+addItemIdToCategory catId itemId = do
+  itemsOrder <- getCategoryItemsOrder catId
+  when (elem itemId itemsOrder) $ throwError ElementIdIsInOrder
```
neongreen: This error is way too vague. What is an "element"? What element was in order?

```haskell
when (elem itemId itemsOrder) $
  throwError $ ItemAlreadyInCategory catId itemId
```

back/src/Guide/Database/Set.hs
```diff
@@ -0,0 +1,570 @@
...
+  let addItemId = itemId : itemsOrder
```
neongreen: Currently new items are added to the end, not to the beginning.

This should not change.

back/src/Guide/Database/Set.hs
```diff
@@ -0,0 +1,570 @@
...
+      encoder = contrazip2 uidParam uidsParam
+      decoder = HD.noResult
+  lift $ HT.statement (catId, newItemsOrder) (Statement sql encoder decoder False)
+
+-- | Add traitId to item, either pros_order or cons_order.
+addTraitIdToItem :: Uid Item -> Uid Trait -> TraitType -> ExceptT DatabaseError Transaction ()
+addTraitIdToItem itemId traitId traitType = do
+    traitsOrder <- getItemTraitsOrder itemId traitType
+    when (elem traitId traitsOrder) $ throwError ElementIdIsInOrder
+    let addTraitId = traitId : traitsOrder
```
neongreen: Same, traits should be added to the end and not to the beginning.

back/src/Guide/Database/Set.hs
```diff
@@ -0,0 +1,570 @@
...
+        WHERE uid = $1
+        |]
+      encoder = contrazip2 uidParam boolParam
+      decoder = HD.noResult
+  lift $ HT.statement (traitId, deleted) (Statement sql encoder decoder False)
+  traitType <- getTraitTypeByTraitId traitId
+  if deleted then deleteTraitIdFromItem itemId traitId traitType
+  else addTraitIdToItem itemId traitId traitType
+
+-- | Move itrait up or down.
```
neongreen:
```suggestion
-- | Move trait up or down.
```

neongreen (approved):

PR #365, opened 2019-08-10
Cleanup the way we work with record fields

neongreen:


back/src/Guide/Database/Types.hs
```diff
@@ -215,8 +221,8 @@ traitToTraitRow
   -> TraitRow
 traitToTraitRow itemId (arg #deleted -> deleted) traitType $(fields 'Trait) =
   TraitRow
-    { traitRowUid = _traitUid
-    , traitRowContent = toText $ show _traitContent
+    { traitRowUid = traitUid
+    , traitRowContent = toText $ show traitContent  -- TODO fix
```
willbasky, line 225: What fix to do?

neongreen: `show` should not be used for business logic when a more precise function exists.

* `show @Int` is ok.
* Using `show` for logging is ok.
* `show @String` or `show @Text` should usually be replaced with more specific functions, e.g. `escapeJSString` or `escapeHTML` or whatever.
* As for `show @MarkdownBlock`, I don't even know off the top of my head what it does. Maybe it pretty-prints the structure, maybe it outputs Markdown, maybe it outputs HTML, I don't know. If you want to write Markdown into the database, use `markdownBlockSource`.

willbasky: let add this extended comment there.

neongreen: It should be added into a styleguide or into a hlint rule or something, not here. I will do that in another PR.

willbasky (approved):

PR #366, opened 2019-08-11
Remove EKG and unused dependencies (found via 'weeder')

neongreen:
TODO: delete Travis cache after this PR is merged because we've gotten rid of perhaps 20-30 deps in total.

### DIFF
--- a/back/guide.cabal
+++ b/back/guide.cabal
@@ -129,12 +129,9 @@ library
                      , filemanip == 0.3.6.*
                      , filepath
                      , fmt
-                     , focus
                      , friendly-time == 0.4.*
                      , hashable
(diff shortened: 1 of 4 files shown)


willbasky (commented):
What if we need ekg monitoring someday?

neongreen:
We likely won't.

For monitoring space/memory usage we have other tools, e.g. DigitalOcean's built-in monitoring. It can also send alerts, which EKG can not do.

For monitoring the number of existing items/categories we can hook into Postgres (or add a graph to Matomo), though I actually don't recall when I wanted to have those stats.

willbasky (approved):
Then we sure.

PR #367, opened 2019-08-11
Make HLint usable

neongreen:
* Reenables and applies some HLint suggestions.
* Suppressed the "reduce duplication" suggestion.
* Adds a hint that forbids -XCPP (which we weren't using anyway).
* Adds a hint that requires certain modules to be imported in a certain way (e.g. you can't write `import qualified Data.Aeson as A`, only `import qualified Data.Aeson as Aeson`).

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,5 +1,32 @@
+# Help Hlint parse our code
 - arguments:
     - -XTypeApplications
+
+# Force qualified imports to be imported consistently.
+- modules:
+    # Common modules from the Haskell ecosystem. The chosen abbreviations
+    # are the most common or almost the most common, as per Hackage stats.
+    # For consistency, Data.ByteString and Data.ByteString.Lazy are
+    # abbreviated as BS and BSL respectively, even though B and BL are
+    # somewhat more common.
+    - { name: Data.ByteString, as: BS }
+    - { name: Data.ByteString.Lazy, as: BSL }
+    - { name: Data.ByteString.Char8, as: BSC }
+    - { name: Data.ByteString.Lazy.Char8, as: BSLC }
+    - { name: Data.Text, as: T }
+    - { name: Data.Text.Lazy, as: TL }
+    - { name: Data.HashMap.Strict, as: HM }
+    - { name: Data.HashMap.Lazy, as: HML }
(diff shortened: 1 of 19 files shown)


neongreen:
* Renames `Edit'Foo` to `EditFoo` because the previous version was kinda weird and because Hlint complained.

neongreen:
This PR is fairly big, you can read just the `.hlint.yaml` changes.

.hlint.yaml
```diff
@@ -1,19 +1,19 @@
...
 - ignore: {name: 'Avoid lambda'}
-- ignore: {name: 'Use camelCase'}
 - ignore: {name: 'Use const'}
 - ignore: {name: 'Use if'}
-- ignore: {name: 'Use notElem'}
 - ignore: {name: 'Use fromMaybe'}
 - ignore: {name: 'Use maybe'}
 - ignore: {name: 'Use fmap'}
 - ignore: {name: 'Use foldl'}
-- ignore: {name: 'Use :'}
```
willbasky, line 16: Does ignoring not enough?

back/src/Guide/Views.hs
```diff
@@ -354,58 +353,58 @@ renderEdit globalState edit = do
...
       let (_, item, trait) = findTrait itemId traitId
       p_ $ "deleted trait from item " >> quote (toHtml (itemName item))
       pre_ $ code_ $ toHtml $ traitContent trait
 
     -- Other
-    Edit'MoveItem itemId direction -> p_ $ do
+    EditMoveItem itemId direction -> p_ $ do
       "moved item " >> printItem itemId
       if direction then " up" else " down"
-    Edit'MoveTrait itemId traitId direction -> do
```
willbasky, line 408: Why `'` sign was used inside names?

willbasky (changes requested):

back/src/Guide/Views.hs
neongreen: Because back when I wrote it, I thought it could be useful to clearly separate the prefix from the meaningful part of the name.

It turned out that it wasn't particularly useful and was somewhat confusing for others.

willbasky (approved):

PR #370, opened 2019-08-12
Implement and use hasql classes

neongreen:


willbasky (approved):
All comments were processed in Slack.

PR #373, opened 2019-08-14
Add deleted field to category

willbasky:


back/src/Guide/Database/Add.hs
```diff
@@ -66,7 +66,7 @@ addCategory
             , enabled_sections
             , items_order
```
neongreen, line 67: Forgot to add `deleted` here.

neongreen (approved):

PR #375, opened 2019-08-14
Don't use explicit imports for Guide.Types.Core

neongreen:


willbasky (changes requested):
If it possible add constrain to hlint for that?

neongreen:
No.

willbasky (approved):

PR #376, opened 2019-08-15
Add CLI

### LINKED ISSUE #374: Improve CLI
Use `optparse-applicative`

Should be following commands:
1. guide
2. guide dry-run
3. guide load-public

And fix travis/makefile

willbasky:
Resolve #374

### DIFF
--- a/back/src/Guide/Cli.hs
+++ b/back/src/Guide/Cli.hs
@@ -14,62 +14,92 @@ import Imports
 
 import Data.Version (showVersion)
 import Development.GitRev (gitCommitDate, gitDirty, gitHash)
-import Options.Applicative (Parser, ParserInfo, command, execParser, fullDesc, help, helper, info,
-                            infoHeader, infoOption, long, metavar, progDesc, short, strOption,
-                            subparser, (<|>))
-import Options.Applicative.Help.Chunk (stringChunk)
+import Options.Applicative (Parser)
 
 import Paths_guide (version)
 
 import qualified Data.Text as T
+import qualified Options.Applicative as Opt
 
 
 -- | All available commands
 data Command
-    = Guide               -- ^ run server
-    | DryRun              -- ^ load database and exit
-    | LoadPublic String   -- ^ load PublicDB, create base on it and exit
+    = RunServer             -- ^ run server
+    | DryRun                -- ^ load database and exit
+    | LoadPublic FilePath   -- ^ load PublicDB, create base on it and exit
 
 ----------------------------------------------------------------------------
 -- Parsers
 ----------------------------------------------------------------------------
 
+-- | Backend uses command line interface:
+{-
+To see help run command:
+$ guide --help
+Usage: guide [-v|--version] [COMMAND]
+
+Available options:
+  -h,--help                Show this help text
+  -v,--version             Show Guide version
+
+Available commands:
+  run                      Run server
+  dry-run                  Load database and exit
+  load-public              Load PublicDB, create base on it and exit
+
+NOTE:
+Command 'guide' is the same as 'guide run'
+
+----------------------------------------------------------------------------
+
+$ guide load-public --help
+Usage: guide load-public (-p|--path FILEPATH)
+  Load PublicDB, create base on it and exit
+
+Available options:
+  -h,--help                Show this help text
+  -p,--path FILEPATH       Public DB file name
+
+-}
+
 -- | Main parser of the application.
 parseCli :: IO Command
-parseCli = execParser
-    $ modifyHeader
-    $ info (helper <*> versionP <*> (pure Guide <|> commandsP))
-    $ fullDesc -- <> progDesc "Guide is wiki"
+parseCli = Opt.execParser
+    $ Opt.info (Opt.helper <*> versionP <*> (pure RunServer <|> commandsP))
+    $ Opt.fullDesc
 
 -- | All possible commands.
 commandsP :: Parser Command
-commandsP = subparser
-    $  command "run" (info (helper <*> pure Guide) $ progDesc "Run server")
-    <> command "dry-run" (info (helper <*> pure DryRun) $ progDesc "Load database and exit")
-    <> command "load-public" (info (helper <*> loadPublic) $ progDesc "Load PublicDB, create base on it and exit")
+commandsP = Opt.subparser
+    $  Opt.command "run" (infoP (pure RunServer) "Start server")
+    <> Opt.command "dry-run" (infoP (pure DryRun) "Load database and exit")
+    <> Opt.command "load-public"
+        (infoP (loadPublic) "Load PublicDB, create base on it and exit")
+  where
+    infoP parser desc = Opt.info (Opt.helper <*> parser) $ Opt.progDesc desc
 
 -- | Parse filepath of PublicDB.
 loadPublic :: Parser Command
-loadPublic = LoadPublic <$> strOption
-    (long "path" <> short 'p' <> help "Public DB file name" <> metavar "FILEPATH")
+loadPublic = LoadPublic <$> Opt.strOption
+    (  Opt.long "path"
+    <> Opt.short 'p'
+    <> Opt.help "Public DB file name"
+    <> Opt.metavar "FILEPATH"
+    )
 
 -- | Parse version option.
 versionP :: Parser (a -> a)
-versionP = infoOption guideVersion
-    $ long "version"
-   <> short 'v'
-   <> help "Show Guide version"
+versionP = Opt.infoOption guideVersion
+    $ Opt.long "version"
+   <> Opt.short 'v'
+   <> Opt.help "Show Guide version"
 
 -- | Show current guide version in git system
 guideVersion :: String
-guideVersion = T.unpack $ T.intercalate "\n" $ [sVersion, sHash, sDate] ++ [sDirty | $(gitDirty)]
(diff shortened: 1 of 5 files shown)


back/src/Guide/Cli.hs
```diff
@@ -0,0 +1,72 @@
...
+
+-- | Parse version option.
+versionP :: Parser (a -> a)
+versionP = infoOption guideVersion
+    $ long "version"
+   <> short 'v'
+   <> help "Show Guide version"
+
+-- | Show current guide version in git system
+guideVersion :: String
```
willbasky, line 99: It is just an option. Can be deleted. This command looks like:
```
$ guide -v
Guide v1.0
 ➤ Git revision: ee8a6e551c2473d4d41f8fd858974a09046b10f7
 ➤ Commit date:  Wed Aug 14 19:42:26 2019 +0000
There are non-committed files.
```

back/src/Guide/Cli.hs
```diff
@@ -0,0 +1,75 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell   #-}
+
+-- | This module adds command line interface to run server.
+--
+-- See help with 'guide --help' to see available commands.
```
neongreen:
```suggestion
-- See help with @guide --help@ to see available commands.
```

This is Haddock so it should use Haddock notation.

back/src/Guide/Cli.hs
```diff
@@ -0,0 +1,75 @@
...
+       ( Command (..)
+       , parseCli
+       ) where
+
+
+import Imports
+
+import Data.Version (showVersion)
+import Development.GitRev (gitCommitDate, gitDirty, gitHash)
+import Options.Applicative (Parser, ParserInfo, command, execParser, fullDesc, help, helper, info,
```
neongreen: It would be better to use a qualified import for this one (`import qualified Options.Applicative as Opt`, for instance). Otherwise this list will have to be amended all the time as additional functions become used and some become unused.

back/src/Guide/Cli.hs
```diff
@@ -0,0 +1,75 @@
...
+                            infoHeader, infoOption, long, metavar, progDesc, short, strOption,
+                            subparser, (<|>))
```
neongreen: `<|>` is already in scope, there's no need to import it. If I'm wrong (maybe I'm wrong), then it should be imported from `Control.Applicative`, not from here.

back/src/Guide/Cli.hs
```diff
@@ -0,0 +1,75 @@
...
+import Options.Applicative.Help.Chunk (stringChunk)
+
+import Paths_guide (version)
+
+import qualified Data.Text as T
+
+
+-- | All available commands
+data Command
+    = Guide               -- ^ run server
```
neongreen: Why not call it `RunServer` then? Or `Start` or something, maybe `Start` is a bit better than `Run`.

back/src/Guide/Cli.hs
```diff
@@ -0,0 +1,75 @@
...
+    | DryRun              -- ^ load database and exit
+    | LoadPublic String   -- ^ load PublicDB, create base on it and exit
```
neongreen: If this is a file path it should say `FilePath`, not `String`.

back/src/Guide/Cli.hs
```diff
@@ -0,0 +1,75 @@
...
+versionP = infoOption guideVersion
+    $ long "version"
+   <> short 'v'
+   <> help "Show Guide version"
+
+-- | Show current guide version in git system
+guideVersion :: String
+guideVersion = T.unpack $ T.intercalate "\n" $ [sVersion, sHash, sDate] ++ [sDirty | $(gitDirty)]
+  where
+    sVersion = "Guide " <> "v" <> T.pack (showVersion version)
```
neongreen:
```suggestion
    sVersion = "Aelve Guide " <> "v" <> T.pack (showVersion version)
```

back/src/Guide/Cli.hs
neongreen: Pretty!

back/src/Guide/Cli.hs
```diff
@@ -0,0 +1,75 @@
...
+    = Guide               -- ^ run server
+    | DryRun              -- ^ load database and exit
+    | LoadPublic String   -- ^ load PublicDB, create base on it and exit
+
+----------------------------------------------------------------------------
+-- Parsers
+----------------------------------------------------------------------------
+
+-- | Main parser of the application.
+parseCli :: IO Command
```
neongreen: Can you post the output of `guide --help`? And, say, `guide load-public --help`?

back/src/Guide/Cli.hs
```diff
@@ -0,0 +1,75 @@
...
+    $ modifyHeader
+    $ info (helper <*> versionP <*> (pure Guide <|> commandsP))
+    $ fullDesc -- <> progDesc "Guide is wiki"
+
+-- | All possible commands.
+commandsP :: Parser Command
+commandsP = subparser
+    $  command "run" (info (helper <*> pure Guide) $ progDesc "Run server")
+    <> command "dry-run" (info (helper <*> pure DryRun) $ progDesc "Load database and exit")
+    <> command "load-public" (info (helper <*> loadPublic) $ progDesc "Load PublicDB, create base on it and exit")
```
neongreen: Please try to keep lines shorter than 80 chars, otherwise working with a split screen is hard.

back/src/Guide/Cli.hs
```diff
@@ -0,0 +1,75 @@
...
+guideVersion = T.unpack $ T.intercalate "\n" $ [sVersion, sHash, sDate] ++ [sDirty | $(gitDirty)]
+  where
+    sVersion = "Guide " <> "v" <> T.pack (showVersion version)
+    sHash = " ➤ " <> ("Git revision: " <> $(gitHash))
+    sDate = " ➤ " <> ("Commit date:  " <> $(gitCommitDate))
+    sDirty = "There are non-committed files."
+
+-- | Put custom header which doesn't cut all spaces
+modifyHeader :: ParserInfo a -> ParserInfo a
+modifyHeader p = p {infoHeader = stringChunk "Guide is wiki"}
```
neongreen: I don't think we need this description (or any kind of description), it will either go out of date quickly or it won't be particularly useful.

back/src/Guide/Main.hs
```diff
@@ -9,12 +9,12 @@
...
 --
---   * Run 'main' to actually start the server.
---   * Run 'mainWith' to run it with a custom config.
+--   * Run 'runGuide' to actually start the server.
+--   * Run 'main' to run it with a custom config.
 module Guide.Main
 (
   main,
-  mainWith,
+  runGuide,
```
neongreen: `dryRun` and `loadPublic` should be exported too.

back/src/Guide/Main.hs
```diff
@@ -9,12 +9,12 @@
 --
 -- This module provides two functions that are of interest:
 --
---   * Run 'main' to actually start the server.
---   * Run 'mainWith' to run it with a custom config.
+--   * Run 'runGuide' to actually start the server.
+--   * Run 'main' to run it with a custom config.
```
neongreen: This seems wrong.

back/src/Guide/Main.hs
```diff
@@ -163,6 +175,26 @@ mainWith config@Config{..} = withLogger config $ \logger -> do
...
+-- | Load database and exit.
+dryRun :: Config -> IO ()
+dryRun config = withLogger config $ \logger -> do
+  db :: DB <- openLocalStateFrom "state/" (error "couldn't load state")
+  logDebugIO logger "loaded the database successfully"
+  closeAcidState db
+  exitSuccess
+
+-- | Load PublicDB, create base on it and exit.
+loadPublic :: String -> Config -> IO ()
```
neongreen:
```suggestion
loadPublic :: FilePath -> Config -> IO ()
```

I also think it should take `Config` as the first argument rather than the second.

back/src/Guide/Main.hs
```diff
@@ -122,34 +121,47 @@ lucidWithConfig x = do
...
+
+-- | Choose command from command line.
+runCommand :: Config -> Command -> IO ()
+runCommand config = \case
+  Guide -> runGuide config
+  DryRun -> dryRun config
+  LoadPublic path -> loadPublic path config
+
+-- | Start the site.
+runGuide :: Config -> IO ()
```
neongreen: Yeah, this should definitely be called `runServer` or `serve` or something, "run Guide" seems too abstract.

back/src/Guide/Cli.hs
willbasky:
```
$ guide --help
Usage: guide [-v|--version] [COMMAND]

Available options:
  -h,--help                Show this help text
  -v,--version             Show Guide version

Available commands:
  run                      Run server
  dry-run                  Load database and exit
  load-public              Load PublicDB, create base on it and exit

NOTE:
Command 'guide' is the same as 'guide run'

$ guide load-public --help
Usage: guide load-public (-p|--path FILEPATH)
  Load PublicDB, create base on it and exit

Available options:
  -h,--help                Show this help text
  -p,--path FILEPATH       Public DB file name
```

neongreen (approved):
Pushed a small commit and approved.

PR #377, opened 2019-08-17
Migrate to hpack

neongreen:
Now we don't have to specify exposed-modules anymore.


back/package.yaml
```diff
@@ -38,61 +38,6 @@ library:
   - DerivingStrategies
   - TemplateHaskellQuotes
   - ScopedTypeVariables
-  exposed-modules:
```
willbasky: No need for them?

willbasky (approved):

PR #378, opened 2019-08-17
Reorganize queries

neongreen:
I've decided to leave the queries sorted by type (insert, select, update, delete) – the alternative is grouping them by the type of object they operate on, but I'm not sure it's actually better.

This PR renames the modules:

* `Guide.Database.Add` -> `Guide.Database.Queries.Insert`
* `Guide.Database.Set` -> `Guide.Database.Queries.Update`
* `Guide.Database.Get` -> `Guide.Database.Queries.Select`

It also creates a new module `Guide.Database.Queries.Delete`.

The logic has not been changed at all. The diff is big because in one module I have changed the order of the functions (trait, item, category -> category, item, trait) for consistency.


neongreen:
Mostly this PR is needed so that the queries would live in their own folder and not mix with stuff like `Guide.Database.Connect`.

I also like that the names correspond to the SQL inside them now.

back/src/Guide/Database/Utils.hs
```diff
@@ -118,7 +118,7 @@ queryRowMaybe = QuasiQuoter
           (#prepared False)
           (#params toPostgresParams)
           (#result (HD.rowMaybe fromPostgresRow))
-          $(stringE sql)|]
+          (fromString $(stringE sql))|]
```
neongreen, line 121: This is needed so that `-XOverloadedStrings` would not be necessary to enable at the use site.

willbasky (approved):

PR #380, opened 2019-08-18
Generate swagger.json for branches

neongreen:


neongreen:
Not tested yet.

back/src/Guide/Cli.hs
```diff
@@ -24,42 +24,10 @@ import qualified Options.Applicative as Opt
...
 data Command
-    = RunServer             -- ^ run server
-    | DryRun                -- ^ load database and exit
-    | LoadPublic FilePath   -- ^ load PublicDB, create base on it and exit
-
-----------------------------------------------------------------------------
--- Parsers
-----------------------------------------------------------------------------
-
-{-
```
neongreen, line 35: I have removed this section because we shouldn't have to update it every time the syntax is changed.

neongreen:
This PR makes it so that there's always an up-to-date `swagger.json` file in the repository. This will make it easier to see whether a PR changes anything in the API, and if so, what exactly.

`swagger.json` isn't the most pleasant thing to read, so in the future I also want to generate a Markdown file with the list of endpoints. (I tried `servant-docs` for it but it wasn't good enough.)

Once it's done, we might be able to get rid of the custom query functions in `ApiSpec` and replace them with `servant-client`. This is yet to be discussed, though.

willbasky (approved):

neongreen:
`[skip ci]` doesn't work because then Mergify is like "oh, but Travis hasn't passed, so I shouldn't merge this".

Okay, let's remove `[skip ci]` and suffer from increased build times.

PR #382, opened 2019-08-18
Migrate to to-1.2.0

neongreen:


neongreen:
I think there won't be any further `to`-related PRs for a while, it covers all conversions we had.

willbasky (approved):

PR #383, opened 2019-08-19
Load into postgres and check it.

willbasky:


willbasky:
> [Tue Aug 20 07:34:12:321331471000 UTC 2019] Debug: <root> | Cat: AcidState == Postgres: False

Why? Reason is in difference of length miliseconds integer between Haskell `UTCTime` and Postgres `timestamptz`. For example:
```
Acid
[ Category 
    { categoryUid = "dyzqbo3k" 
    , categoryTitle = "cabal" 
    , categoryCreated = 2019-08-20 04:57:25.063201537 UTC
...
```
and
```
Postgres
[ Category 
    { categoryUid = "dyzqbo3k" 
    , categoryTitle = "cabal" 
    , categoryCreated = 2019-08-20 04:57:25.063201 UTC
```

back/src/Guide/Database/Queries/Update.hs
```diff
@@ -95,7 +95,7 @@ updateCategoryRow catId f = do
   when (old_categoryRowItemsOrder /= new_categoryRowItemsOrder) $ do
     let statement :: Statement (Uid Category, [Uid Item]) ()
         statement = [execute|UPDATE categories SET items_order = $2 WHERE uid = $1|]
-    lift $ HT.statement (catId, new_categoryRowItemsOrder) statement
+    lift $ HT.statement (catId, nub new_categoryRowItemsOrder) statement
```
willbasky, line 98: To exclude id duplicates.
It happens when adding category with full CategoryRow converted with `categoryToCategoryRow` and then adding items. In that case item order will be updated two times. 
Both `categoryToCategoryRow` and `insertItem` updates itemOrder.

.travis.yml
```diff
@@ -98,6 +100,17 @@ jobs:
             >&2 echo "FAIL: found $undocumented top-level undocumented definitions. The current limit is 251. Please, add Haddock comments to undocumented definitions!"
             exit 1
           fi
+        # Load AcidState to Postgres and check if they are equal
+        - stack exec --cwd back -- guide load-into-postgres 2> postgres.log
+        - cat postgres.log
```
neongreen: This is more complicated than needed.

If you want to check if a file contains a specific string, you can do:

```bash
if grep -q True postgres.log; then
  ...
else
  ...
fi
```

But it can be even simpler – `guide load-into-postgres` should just do `exitFailure` if the databases are not equal. 

Parsing text output is in general a kinda fragile thing and you shouldn't do it if you can avoid it.

Makefile
```diff
@@ -35,6 +35,13 @@ back/test-db: back
 back/run:
 	stack exec --cwd back -- guide
 
+# Load AcidState base to Postgres and check equality.
+# Postgres should be installed already.
+# 'Guide' base should be created already and clear.
+.PHONY: back/postgres
+back/postgres:
```
neongreen: Calling it `back/load-into-postgres` would be better.

back/src/Guide/Database/Migration.hs
```diff
@@ -0,0 +1,186 @@
+-- | Module contains all stuff to migrate from AcidState to Postgres.
+module Guide.Database.Migration
```
neongreen: This is not a good module name because we also have Postgres migrations – it's easy to mix them up. `Guide.Database.Import` would be better.

back/src/Guide/Database/Migration.hs
```diff
@@ -0,0 +1,186 @@
...
+
+
+-- | Load categories and deleted categories from acid state to postgres
+-- and check if they are equal.
+--
+-- NOTE: It loads categories and categoriesDeleted fields of GlobalState only.
+loadIntoPostgres :: Config -> IO ()
+loadIntoPostgres config@Config{..} = withLogger config $ \logger -> do
+  withDB (pure ()) $ \db -> do
+    globalState@GlobalState{..} <- dbQuery logger db GetGlobalState
```
neongreen:
```suggestion
    globalState <- dbQuery logger db GetGlobalState
```

neongreen: Well, or if you want to specify the type, say `globalState :: GlobalState <- ...`.

back/src/Guide/Types/Core.hs
```diff
@@ -281,7 +281,10 @@ data Category = Category {
   -- 'ItemNotesSection', then notes will be shown for each item
   categoryEnabledSections :: Set ItemSection
   }
-  deriving (Show, Generic, Data)
+  deriving (Generic, Data, Eq, Show)
+
+instance Ord Category where
```
neongreen: This is not a good instance to have – it's surprising that `compare cat1 cat2 == EQ` does not imply `cat1 == cat2`, and it can trip someone up. Why do you need it?

back/src/Guide/Database/Utils.hs
```diff
@@ -140,6 +140,7 @@ queryRows = QuasiQuoter
 --
 -- Like 'queryRowMaybe', but returns @Statement a ()@ instead of @Statement
 -- a (Maybe b)@.
+-- NOTE: 'execute' consumes only one query at time.
```
neongreen, line 143: Did you check that there are no other buggy queries containing several statements?

back/src/Guide/Database/Queries/Update.hs
```diff
@@ -216,9 +221,9 @@ updateTraitRow
   :: Uid Trait
   -> (TraitRow -> TraitRow)
   -> ExceptT DatabaseError Transaction ()
-updateTraitRow catId f = do
+updateTraitRow traitId f = do
```
neongreen, line 224: 👍

.travis.yml
neongreen: Also, this step belongs in the "test the backend" job, not "build the backend".

neongreen: Also, since you added a makefile command you can use it here instead of doing `stack exec`.

back/src/Guide/Database/Migration.hs
```diff
@@ -0,0 +1,186 @@
...
+postgresLoader logger globalState@GlobalState{..} = do
+    -- Postgres should be started and 'guide' base created.
+    setupDatabase
+
+    -- Upload to Postgres
+    conn <- connect
+    runTransactionExceptT conn Write $ insertCategories globalState
+
+    -- Download from Postgres
+    (catPosg, catDelPosg) <- runTransactionExceptT conn Read getCategories
```
neongreen: Please don't abbreviate "Postgres" and "Deleted", it's hard to understand these names.

back/src/Guide/Database/Migration.hs
```diff
@@ -0,0 +1,186 @@
...
+
+    -- Prepare data
+    let catAcidNorm = map normalizeUTC categories
+    let catAcidDelNorm = map normalizeUTC categoriesDeleted
+
+    -- Check equality of availiable categories
+    let catLength = length catPosg == length catAcidNorm
```
neongreen: Checking the length is not necessary.

back/src/Guide/Database/Migration.hs
```diff
@@ -0,0 +1,186 @@
...
+    let cats = sort catPosg == sort catAcidNorm
```
neongreen: You can `sortOn categoryUid`, no need for the `Ord Category` instance.

back/src/Guide/Database/Migration.hs
```diff
@@ -0,0 +1,186 @@
...
+-- | Read something from the database.
+dbQuery :: (EventState event ~ GlobalState, QueryEvent event, Show event)
+        => Logger -> DB -> event -> IO (EventResult event)
+dbQuery logger db x = do
+  logDebugIO logger $ "dbQuery: " +|| x ||+ ""
+  liftIO $ query db x
+
+-- normalizeUTC :: GlobalState -> GlobalState
+normalizeUTC :: Category -> Category
+normalizeUTC = transformBis [[transformer cutUTCTime]]
```
neongreen: What is `transformBis` for? Why not `transformBi`?

```suggestion
normalizeUTC = transformBi cutUTCTime
```

back/src/Guide/Database/Migration.hs
```diff
@@ -0,0 +1,186 @@
...
+
+cutUTCTime :: UTCTime -> UTCTime
```
neongreen: There should be a comment explaining why we do it.

back/src/Guide/Database/Migration.hs
```diff
@@ -0,0 +1,186 @@
...
+cutUTCTime UTCTime{..} = UTCTime{utctDay, utctDayTime = utctDayTimeCut}
+  where
+    utctDayTimeCut = picosecondsToDiffTime pico12Cut
+    pico12 = diffTimeToPicoseconds utctDayTime
+    pico12Cut = floor ((fromInteger pico12 / 1000000) :: Double) * 1000000
```
neongreen: It would be better to truncate the categories before writing them into the database, not after. This way we wouldn't have to wonder if Postgres needs `floor` or `ceiling` or `round` or whatever.

back/src/Guide/Database/Migration.hs
```diff
@@ -0,0 +1,186 @@
...
+
+----------------------------------------------------------------------------
+-- Insert helpers
+----------------------------------------------------------------------------
+
+-- | Insert all categories from AcidState either deleted or not.
+insertCategories :: GlobalState -> ExceptT DatabaseError Transaction ()
+insertCategories GlobalState{..} = do
+  mapM_ (insertCategoryWhole (#deleted False)) categories
```
neongreen: You can just inline these into `postgresLoader`, since the `insertCategories` function is kinda non-reusable anyway.

back/src/Guide/Database/Queries/Update.hs
neongreen: `categoryToCategoryRow` does not exist, are you talking about some other function?

back/src/Guide/Database/Queries/Update.hs
```diff
@@ -181,6 +181,11 @@ updateItemRow itemId f = do
     let statement :: Statement (Uid Item, Bool) ()
         statement = [execute|UPDATE items SET deleted = $2 WHERE uid = $1|]
     lift $ HT.statement (itemId, new_itemRowDeleted) statement
+    if new_itemRowDeleted
```
neongreen, line 184: I have to say that I don't like how these functions are becoming too smart. It's not `updateItemRow`'s job to maintain database invariants. Something like `_categoryRowItemsOrder %~ (++ [itemId])` belongs in the business logic, not here – for instance, we would definitely want to insert the item ID somewhere else, not to the end of the category. (When restoring a deleted item we record its position and then restore back to that position.)

I think database queries should in general be pretty dumb and all logic should belong in the functions using those queries.

back/src/Guide/Database/Queries/Update.hs
willbasky: yes, `categoryToRowCategory`

Makefile
willbasky: it is too long, though it's accurate.

back/src/Guide/Database/Migration.hs
willbasky: just have forgotten to delete.

back/src/Guide/Types/Core.hs
willbasky: It just to sort categories by Uid, otherwise AcidState has one order, Postgres has another one and can't be checked for identity.
```
> [1,2,3] == [1,3,2]
False
> [1,2,3] == Data.List.sort [1,3,2]
True
```

back/src/Guide/Database/Utils.hs
willbasky: Have checked just now, all is clear.

back/src/Guide/Database/Migration.hs
willbasky: There it is!

back/src/Guide/Database/Queries/Update.hs
willbasky: Next commit refactor all but did not this one above.

neongreen (approved):

PR #390, opened 2019-08-29
#389 jsonb

### LINKED ISSUE #389: Change Postgres to store data in JSON
Tables (categories, items, traits) should be compressed into a single `categories` table:

```sql
    CREATE TABLE categories (
      uid text PRIMARY KEY,           -- Unique category ID
      data jsonb NOT NULL
    );
```

- [x] Change the schema. (No need to write a migration, just change the v0 schema.)
- [x] Implement `selectCategories`, `selectCategory`, `insertCategory`, `updateCategory`.
- [ ] Benchmark how long it takes to execute `selectCategories` with code compiled with `-O` on the official database.
- [ ] Benchmark how long it takes to do `updateCategory id` on the biggest category with code compiled with `-O` on the official database.

Based on the latter two points we can decide whether we need any fine-grained select/update functions or whether we can get away with simply reading and writing everything all the time.

willbasky:
Resolve #389

---
This change is [Reviewable](https://reviewable.io/reviews/aelve/guide/390)


back/benchmarks/Main.hs
```diff
@@ -0,0 +1,49 @@
+-- | Module contains all stuff to migrate from AcidState to Postgres.
```
neongreen: Wrong description?

back/benchmarks/Main.hs
```diff
@@ -0,0 +1,49 @@
...
+main = do
+    conn <- connect
+    defaultMain [databaseBenchmark conn]
+  where
+    databaseBenchmark conn =
+      bgroup
+        "Database"
+        [ bench "select" $ nfIO $
+            runTransactionExceptT conn Read $ selectCategory "category1111"
+        , bench "updete" $ nfIO $
```
neongreen:
```suggestion
        , bench "update" $ nfIO $
```

back/benchmarks/Main.hs
```diff
@@ -0,0 +1,49 @@
...
+main :: IO ()
+main = do
+    conn <- connect
+    defaultMain [databaseBenchmark conn]
+  where
+    databaseBenchmark conn =
+      bgroup
+        "Database"
+        [ bench "select" $ nfIO $
+            runTransactionExceptT conn Read $ selectCategory "category1111"
```
neongreen: Please let's not have such names in code, they look sloppy. "category" or "uid1" would be better.

back/benchmarks/Main.hs
```diff
@@ -0,0 +1,49 @@
...
+
+import Guide.Database.Queries.Update
+import Guide.Database.Queries.Select
+import Guide.Types.Core
+import Guide.Database.Connection
+
+
+main :: IO ()
+main = do
+    conn <- connect
```
neongreen: We have two-space indentation.

back/benchmarks/Main.hs
```diff
@@ -0,0 +1,49 @@
...
+    databaseBenchmark conn =
+      bgroup
+        "Database"
+        [ bench "select" $ nfIO $
+            runTransactionExceptT conn Read $ selectCategory "category1111"
+        , bench "updete" $ nfIO $
+            runTransactionExceptT conn Write $ updateCategory "category1111" update
+        ]
+    update :: Category -> Category
+    update = _categoryTitle .~ "title10"
```
neongreen: After the first application of `update` the title will be `"title10"`, so `update` will be equivalent to `id` and `updateCategory` will not do any writes. You should do something that would change the category _every_ time, e.g. `_categoryTitle <>~ "+"`, and then re-run the benchmarks.

back/benchmarks/Main.hs
```diff
@@ -0,0 +1,49 @@
...
+
+{-
+benchmarked Database/select
+time                 496.1 μs   (429.1 μs .. 551.4 μs)
+                     0.932 R²   (0.868 R² .. 0.976 R²)
+mean                 590.7 μs   (502.5 μs .. 939.6 μs)
+std dev              508.6 μs   (51.75 μs .. 1.065 ms)
+variance introduced by outliers: 97% (severely inflated)
```
neongreen: You should run benchmarks without anything running in background (browsers, etc), otherwise they will be imprecise. "severely inflated" is not good.

back/package.yaml
```diff
@@ -60,6 +60,8 @@ library:
   # You don't need to add modules here, all modules will be exposed automatically.
   #
   # exposed-modules:
+  ghc-options:
```
neongreen: `stack` already compiles with `-O` when you don't pass `--fast`.

back/package.yaml
```diff
@@ -197,3 +200,17 @@ tests:
     - temporary
     - webdriver
     - yaml
+
+benchmarks:
```
neongreen, line 201: You also should add `--enable-benchmarks` (or `--bench --no-run-benchmarks` or something, look at `stack --help`) in the Travis config so that they would be compiled during CI.

back/src/Guide/Uid.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE DeriveAnyClass    #-}
```
neongreen: I think this is unnecessary.

back/src/Guide/Database/Queries/Insert.hs
```diff
@@ -100,6 +103,51 @@ insertCategoryWithCategoryRow categoryRow = do
         |]
   lift $ HT.statement categoryRow statement
 
+insertCategory2
```
neongreen: You can comment out the old code instead of leaving it around and adding `2` to the new functions. Since the old schemas are commented out, I think it's alright to also comment out the old updates/selects/etc.

back/src/Guide/Database/Queries/Select.hs
```diff
@@ -46,6 +48,24 @@ import Guide.Uid
...
+
+selectCategory :: Uid Category -> ExceptT DatabaseError Transaction Category
+selectCategory catId = do
+  mCatRow <- selectCategoryMaybe catId
+  case mCatRow of
+    Nothing     -> throwError $ CategoryNotFound catId
+    Just catRow -> pure catRow
+
 -- | Get a 'CategoryRow'.
 selectCategoryRowMaybe :: Uid Category -> ExceptT DatabaseError Transaction (Maybe CategoryRow)
```
neongreen: This code should also be commented out.

back/src/Guide/Database/Schema.hs
```diff
@@ -103,86 +103,95 @@ setSchemaVersion version = do
...
-      notes text NOT NULL,            -- Category notes as Markdown
-      enabled_sections text[]         -- Item sections to show to users; the list of
-        NOT NULL,                     --   possible section names is defined by backend
-      items_order text[]              -- Uids of items in the category; this list
-        NOT NULL,                     --   specifies in what order they should be
-                                      --   displayed, and is necessary to allow moving
-                                      --   items up and down
-      deleted boolean                 -- Whether the category is deleted
-        NOT NULL
+      data jsonb NOT NULL        -- Single category with all items and raits belong to it
```
neongreen:
```suggestion
      data jsonb NOT NULL        -- Single category with all items and traits belong to it
```

back/src/Guide/Markdown.hs
```diff
@@ -46,38 +47,58 @@ import ShortcutLinks.All (hackage)
...
+deriving instance NFData MD.NodeType
+deriving instance NFData MD.ListAttributes
+deriving instance NFData MD.DelimType
+deriving instance NFData MD.ListType
+deriving instance NFData MD.PosInfo
+deriving instance NFData (WithSource [MD.Node])
+deriving instance (NFData b, NFData t) => NFData (Document t b)
+deriving instance (NFData b, NFData t) => NFData (Section t b)
+deriving instance NFData Heading
+-- instance NFData (Node a f) where
```
neongreen: Why is this comment here?

back/src/Guide/Markdown.hs
```diff
@@ -320,6 +341,19 @@ instance Aeson.ToJSON MarkdownTree where
...
+    txt <- o Aeson..: "text"
+    pure $ toMarkdownInline txt
+instance Aeson.FromJSON MarkdownBlock where
+  parseJSON = Aeson.withObject "MarkdownBlock" $ \o -> do
+    txt <- o Aeson..: "text"
+    pure $ toMarkdownBlock txt
+instance Aeson.FromJSON MarkdownTree where
+  parseJSON = Aeson.withObject "MarkdownTree" $ \o -> do
+    txt <- o Aeson..: "text"
+    pure $ toMarkdownTree "" txt
```
neongreen: You should also get `markdownTreeIdPrefix`. This is an omission in the `ToJSON` instance I wrote, it should be there as well.

back/src/Guide/Markdown.hs
```diff
@@ -362,3 +396,21 @@ instance SafeCopy MarkdownTree where
...
     toMarkdownTree <$> safeGet <*> safeGet
+
+instance ToPostgres MarkdownInline where
+  toPostgres = markdownInlineSource >$< HE.text
+
+instance ToPostgres MarkdownBlock where
+  toPostgres = markdownBlockSource >$< HE.text
+
+instance ToPostgres MarkdownTree where
+  toPostgres = markdownTreeSource >$< HE.text
```
neongreen: Also the prefix here. You can use a composite Postgres type.

back/src/Guide/Markdown.hs
```diff
@@ -362,3 +396,21 @@ instance SafeCopy MarkdownTree where
...
+
+instance FromPostgres MarkdownInline where
+  fromPostgres = toMarkdownInline <$> HD.text
+
+instance FromPostgres MarkdownBlock where
+  fromPostgres = toMarkdownBlock <$> HD.text
+
+instance FromPostgres MarkdownTree where
+  fromPostgres = toMarkdownTree "" <$> HD.text
```
neongreen: And get back the prefix here.

back/src/Guide/Types/Core.hs
```diff
@@ -74,6 +76,20 @@ instance Aeson.ToJSON Trait where
   toJSON = Aeson.genericToJSON Aeson.defaultOptions {
     Aeson.fieldLabelModifier = over _head toLower . drop (T.length "trait") }
 
+instance Aeson.FromJSON Trait where
```
neongreen, line 80: If this instance is not using `generic`, the `ToJSON` instance should also not use `generic`.

back/src/Guide/Types/Core.hs
```diff
@@ -74,6 +76,20 @@ instance Aeson.ToJSON Trait where
...
+  parseJSON = Aeson.withObject "Trait" $ \o -> do
+    traitUid     <- o Aeson..: "uid"
+    content      <- o Aeson..: "content"
```
neongreen: Since `MarkdownInline` has a `FromJSON` instance, you can just write `traitContent <- o Aeson..: "content"`.

back/src/Guide/Types/Core.hs
```diff
@@ -74,6 +76,20 @@ instance Aeson.ToJSON Trait where
...
+    traitContent <- toMarkdownInline <$> content Aeson..: "text"
+    pure Trait{..}
+
+instance ToPostgres Trait where
```
neongreen: Is this instance used anywhere?

back/src/Guide/Types/Core.hs
```diff
@@ -74,6 +76,20 @@ instance Aeson.ToJSON Trait where
...
+  toPostgres = toByteString . Aeson.encode >$< HE.jsonbBytes
```
neongreen: Why are you using `jsonbBytes` instead of just `jsonb`? (Same for all other instances.)

back/src/Guide/Types/Core.hs
```diff
@@ -216,6 +232,33 @@ instance Aeson.ToJSON Item where
...
     Aeson.fieldLabelModifier = over _head toLower . drop (T.length "item") }
 
+instance Aeson.FromJSON Item where
+  parseJSON = Aeson.withObject "Item" $ \o -> do
+    itemUid         <- o Aeson..: "uid"
+    itemName        <- o Aeson..: "name"
+    itemCreated     <- o Aeson..: "created"
+    itemHackage     <- o Aeson..:? "hackage"
+    summary         <- o Aeson..: "summary"
+    itemSummary     <- toMarkdownBlock <$> summary Aeson..: "text"
```
neongreen: Again, `toMarkdownBlock` is not needed because MarkdownBlock already has a FromJSON instance.

back/benchmarks/Main.hs
willbasky: length "category1111" == 12 as Uid needed.

back/benchmarks/Main.hs
willbasky: ok. then speed is
Screenshot_20190901_182126

back/benchmarks/Main.hs
willbasky: I reduced inflated parameter to 53% and 39%

back/package.yaml
```diff
@@ -197,3 +197,17 @@ tests:
...
+    source-dirs: benchmarks
+    dependencies:
+      - base <5
+      - guide
+      - gauge
+      - hasql
+      - hasql-transaction
+
+    ghc-options:
+    - -O
```
neongreen: I think this won't have any effect if the library is not compiled with `-O`, so you can just remove it and add a short instruction on how to run benchmarks properly to the `README`.

back/src/Guide/Database/Import.hs
```diff
@@ -20,12 +20,11 @@ import Guide.Database.Schema
 import Guide.Database.Types
 import Guide.State
 import Guide.Types.Core
-import Guide.Uid
 import Guide.Config
 import Guide.Logger
 
 
--- | Load categories and deleted categories from acid state to postgres
+-- | Load categories and archives categories from acid state to postgres
```
neongreen:
```suggestion
-- | Load categories and archived categories from acid state to postgres
```

Is this what you meant?

back/src/Guide/Database/Import.hs
```diff
@@ -38,33 +37,42 @@ loadIntoPostgres config@Config{..} = withLogger config $ \logger -> do
...
     setupDatabase
     -- Upload to Postgres
     conn <- connect
     runTransactionExceptT conn Write $ insertCategories globalState
     -- Download from Postgres
-    (catPostgres, catDeletedPostgres)
-      <- runTransactionExceptT conn Read getCategories
+    catPostgres <- runTransactionExceptT conn Read $
+      selectCategories (#archived False)
+    catarchivedPostgres <- runTransactionExceptT conn Read $
```
neongreen:
```suggestion
    catArchivedPostgres <- runTransactionExceptT conn Read $
```

back/src/Guide/Database/Import.hs
```diff
@@ -38,33 +37,42 @@ loadIntoPostgres config@Config{..} = withLogger config $ \logger -> do
...
+      selectCategories (#archived True)
     -- Check identity of available categories
     let checkedCat =
           sortOn categoryUid catPostgres ==
           sortOn categoryUid (categories globalState)
-    -- Check identity of deleted categories
+    -- Check identity of archived categories
     let checkedCatDeleted =
-          sortOn categoryUid catDeletedPostgres ==
+          sortOn categoryUid catarchivedPostgres ==
```
neongreen:
```suggestion
          sortOn categoryUid catArchivedPostgres ==
```

back/src/Guide/Database/Queries/Update.hs
```diff
@@ -21,13 +23,44 @@ import Guide.Database.Types
...
+          WHERE uid = $1|]
+    lift $ HT.statement (catId, new_category) statement
+
+-- | Update category archived field when it is different then passed parameter.
+updateCategoryArchived
+  :: Uid Category
+  -> "archived" :! Bool
+  -> ExceptT DatabaseError Transaction ()
+updateCategoryArchived catId (arg #archived -> archived) = do
+  isArchived <- isCategoryArchived catId
```
neongreen: "old_archived" and "new_archived" would be more consistent.

back/src/Guide/Markdown.hs
```diff
@@ -318,7 +330,22 @@ instance Aeson.ToJSON MarkdownBlock where
...
+    txt <- o Aeson..: "text"
+    pure $ toMarkdownInline txt
+instance Aeson.FromJSON MarkdownBlock where
+  parseJSON = Aeson.withObject "MarkdownBlock" $ \o -> do
+    txt <- o Aeson..: "text"
+    pure $ toMarkdownBlock txt
+instance Aeson.FromJSON MarkdownTree where
+  parseJSON = Aeson.withObject "MarkdownTree" $ \o -> do
+    txt <- o Aeson..: "text"
+    prefix <- o Aeson..:? "prefix" Aeson..!= T.empty
```
neongreen: You can use `""` instead of `T.empty`, I don't think there is a reason to not use overloaded strings here.

back/src/Guide/Types/Core.hs
```diff
@@ -316,8 +353,44 @@ changelog ''Category (Past 9, Past 8) []
...
+instance ToPostgres Category where
+  toPostgres = Aeson.toJSON >$< HE.jsonb
+
+instance FromPostgres Category where
+  fromPostgres = resultToEither . Aeson.fromJSON <$> HD.jsonb
+
+-- | Unwarp result to category or fail.
+resultToEither :: Aeson.Result Category -> Category
+resultToEither (Aeson.Success category) = category
+resultToEither (Aeson.Error s) = error $ "fromJSON failed with error: " ++ s
```
neongreen: No, `error` is verboten. Apparently we'll have to use `jsonbBytes` after all because it would let us give back the decoding error (as `Left`) without using `error`.

PR #396, opened 2019-09-09
Commit the JS bundle for the old frontend into the repo

neongreen:


willbasky (approved):
We are close to simplicity.

PR #408, opened 2019-10-31
Flag to ignore disabled sections

### LINKED ISSUE #337: Don't return info from disabled sections
Might require coordination with the frontend.

willbasky:
Resolves #337

---
This change is [Reviewable](https://reviewable.io/reviews/aelve/guide/408)


back/src/Guide/Api/Types.hs
```diff
@@ -153,6 +153,10 @@ data ItemSite route = ItemSite
   { _getItem :: route :-
       Summary "Get item by id"
       :> ErrorResponse 404 "Item not found"
+      :> "ignore selections"
```
willbasky: Does the space is bad?

